PR #2, opened 2020-07-08
subscription iterators

aricart:
- Changed subscriptions to use iterators
- Made it possible to still specify a callback should it be necessary by setting the callback function as an option in the subscription
- See tests/iterators_test.ts for an example of what is possible


aricart:
@kozlovic skip reviewing all the changes in the tests to accommodate this feature, except for tests/iterators_test.ts.

nats-base-client/types.ts
```diff
@@ -117,14 +118,14 @@ export interface Req extends Base {
   token: string;
 }
 
-export interface Sub extends Base {
-  sid: number;
-  queue?: string | null;
-}
+// export interface Sub extends Base {
```
kozlovic, line 121: Should you remove?

nats-base-client/types.ts
```diff
@@ -117,14 +118,14 @@ export interface Req extends Base {
...
+//   sid: number;
+//   queue?: string | null;
+// }
 
-export function defaultSub(): Sub {
-  return { sid: 0, subject: "", received: 0 } as Sub;
-}
+// export function defaultSub(): Sub {
```
kozlovic, line 126: same

tests/iterators_test.ts
```diff
@@ -0,0 +1,122 @@
+import { connect } from "../src/connect.ts";
```
kozlovic, line 1: copyright banner missing

PR #8, opened 2020-07-22
Headers

aricart:
feature - Implemented header support for ES clients (deno and ws)


nats-base-client/util.ts
```diff
@@ -200,3 +202,31 @@ export function shuffle(a: any[]): any[] {
...
+    raw = raw.slice(1);
+  }
+  for (const rh of raw) {
+    const sep = rh.indexOf(":");
+    if (sep < 0) {
+      continue;
+    }
+    const k = rh.slice(0, sep);
+    const v = rh.slice(sep + 1);
+    headers.set(k, v);
```
kozlovic, line 229: It is allowed to have multiple times the same keys, in which case, the map entry should contain an array of strings, not just single value. For instance in Go you have the Headers.Add() which adds as opposed to Set() which replaces.

nats-base-client/util.ts
```diff
@@ -200,3 +202,31 @@ export function shuffle(a: any[]): any[] {
   }
   return a;
 }
+
+export function encodeHeader(headers: Headers): Uint8Array {
+  let v = `NATS/1.0\r\n`;
+  for (const [key, value] of headers) {
+    v += `${key}: ${value}\r\n`;
```
kozlovic, line 209: See comment about decode. You may have to repeat the same key with a different value.
Also, the server generally does not check headers content, but we already have a case where it will check for a Msg-Id header for JetStream de-dup feature. Since it does a byte scan of this header, it will not be tolerant of the casing which it assumes will be canonical ("my-key" becomes "My-Key"). Libraries are therefore expected to send the headers in a canonical form. (see thread here: https://github.com/nats-io/nats-server/pull/1508#discussion_r452382516)

nats-base-client/headers.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  static decode(a: Uint8Array): MsgHeaders {
+    const mh = new MsgHeaders();
+    const s = new TextDecoder().decode(a);
+    const lines = s.split("\r\n");
+    const h = lines[0];
+    if (h !== MsgHeaders.HEADER) {
+      const str = h.replace(MsgHeaders.HEADER, "");
+      mh.error = parseInt(str, 10);
```
kozlovic, line 74: Not sure I understand this code, but are you parsing the status if it is sent like this: `NATS/1.0 503\r\n` here? If not, know that it will be coming in another feature (the no responders) where we may get this kind of header).

nats-base-client/headers.ts
```diff
@@ -0,0 +1,194 @@
...
+    } else {
+      lines.slice(1).map((s) => {
+        if (s) {
+          const [k, v] = s.split(":");
+          mh.append(k, v);
```
kozlovic, line 79: I think you should trim left (not sure about right) the value.

nats-base-client/headers.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  toString(): string {
+    if (this.headers.size === 0) {
+      return "";
+    }
+    let s = MsgHeaders.HEADER;
+    for (const [k, v] of this.headers) {
+      for (let i = 0; i < v.length; i++) {
+        s = `${s}\r\n${k}:${v[i]}`;
```
kozlovic, line 93: By convention, a space is usually added after the `:`. That may be important if you plan to compare incoming header with a "toString()" version?

kozlovic: Which incidentally you are doing in "encodeHeader"

nats-base-client/util.ts
```diff
@@ -200,3 +202,37 @@ export function shuffle(a: any[]): any[] {
...
+}
+
+export function decodeHeaders(buf: Uint8Array): Headers {
+  const headers = new Headers();
+  const v = new TextDecoder("utf-8").decode(buf);
+  let raw = v.split("\r\n");
+  let proto = raw[0];
+  const code = proto.replace("NATS/1.0 ", "");
+  if (code !== "") {
+    // this is only currently a timeout
```
kozlovic, line 222: I don't understand what you are doing here.

nats-base-client/util.ts
kozlovic: Still applies, you should call add here instead. Also, should trim the value.

tests/msgheaders_test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  let header = MsgHeaders.canonicalMIMEHeaderKey("foo");
+  assertEquals("Foo", header);
+  assert(h.has("Foo"));
+  assert(h.has("foo"));
+  const foos = h.values(header);
+  assertEquals(2, foos.length);
+  assert(foos.indexOf("bar") !== -1);
+  assert(foos.indexOf("bam") !== -1);
+
```
kozlovic, line 45: Just to verify I would test that indexOf("bat") does return -1.

nats-base-client/headers.ts
aricart: yes this is wip - next commit will answer it - as it will fail the request with an error.

nats-base-client/headers.ts
aricart: This is mostly non-sensical issue - the minute that space is allowed in the value (it is), the value next to the colon, cannot be optional, because AFAIK, it is the user value. The mime specification again is not the right guidance for this - I am OK with writing the go version that the server and client use, we can eliminate a lot of ambiguity and treat the keys and values in the headers as they should be.

nats-base-client/util.ts
aricart: parsing the error - removed the NATS/1.0, and if it is not empty, parsing a number out.

nats-base-client/headers.ts
kozlovic: From https://tools.ietf.org/html/rfc7230#section-3.2.4:
```
A field value might be preceded and/or followed by optional
whitespace (OWS); a single SP preceding the field-value is preferred
for consistent readability by humans.  The field value does not
include any leading or trailing whitespace: OWS occurring before the
first non-whitespace octet of the field value or after the last
non-whitespace octet of the field value ought to be excluded by
parsers when extracting the field value from a header field.
```
and again, this is what you are doing in the encoding of headers: https://github.com/nats-io/nats.deno/pull/8/files#diff-1c08cf760d969e81eb7b6018e0cc6c80R209

nats-base-client/headers.ts
kozlovic: You are going against uphill battle here since Go client will send the "k: v" and so your library will report that the value is " v" instead of "v"?

nats-base-client/headers.ts
aricart: https://github.com/nats-io/nats.deno/pull/8/files#diff-812a477358fb5cc1c60403e0b9fe3b35R301 is where this is looked at.

nats-base-client/headers.ts
aricart: on add, the value checker will now do a `trimLeft()`

kozlovic: Can you point this to me? I did not see that.

nats-base-client/headers.ts
aricart: 👍

tests/msgheaders_test.ts
aricart: for some reason, this is not updating here:
```javascript
const foos = h.values(header);
  assertEquals(2, foos.length);
  assertArrayContains(foos, ["bar", "bam"]);
  assert(foos.indexOf("baz") === -1);
```

kozlovic: No, I saw you added that.

nats-base-client/headers.ts
```diff
@@ -64,18 +74,21 @@ export class MsgHeaders implements Headers {
...
-    if (h !== MsgHeaders.HEADER) {
-      const str = h.replace(MsgHeaders.HEADER, "");
+    if (h !== NatsHeaders.HEADER) {
+      const str = h.replace(NatsHeaders.HEADER, "");
       mh.error = parseInt(str, 10);
     } else {
       lines.slice(1).map((s) => {
         if (s) {
-          const [k, v] = s.split(":");
+          //
```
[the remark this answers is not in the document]
aricart: this also didn't update.

nats-base-client/headers.ts
aricart: https://github.com/nats-io/nats.deno/pull/8/files#diff-e2d29cee7d766c157d9cc8d08fd9955fR167

PR #12, opened 2020-07-31
changes to the request api

aricart:
- [BREAKING] refactored request API to match publish API
- `timeout` is now an option `request(subject, data, {timeout: 1000, headers: h})`
- [FEAT] added the ability to make requests that specify headers.


README.md
```diff
@@ -289,7 +289,7 @@ const nc = await connect({ url: "demo.nats.io:4222" });
 // a client makes a request and receives a promise for a message
 // by default the request times out after 1s (1000 millis) and has
 // no payload.
-await nc.request("time", 1000, "hello!")
+await nc.request("time", "hello!", { timeout: 1000 })
```
kozlovic, line 292: Why though? Meaning you did not address that in the summary of the PR: why is timeout an option now? Requesting without a timeout sounds to me like a publish.

nats-base-client/types.ts
```diff
@@ -59,18 +59,17 @@ export enum Payload {
 }
 
 export interface ConnectionOptions {
+  authenticator?: Authenticator;
   debug?: boolean;
   headers?: boolean;
-  maxPingOut?: number;
```
kozlovic, line 64: It happens very often in your PRs that fields like that (or errors) move location in that list. Is that an automatic re-ordering from your IDE or something? It is very distracting and always have to look if it has been completely removed or just moved, etc..

src/deno_transport.ts
```diff
@@ -93,14 +93,13 @@ export class DenoTransport implements Transport {
...
-            new Error("socket closed while expecting INFO"),
-          );
-        }
+      const c = await this.conn.read(this.buf);
+      if (c === null) {
+        // EOF
+        return Promise.reject(
+          new Error("socket closed while expecting INFO"),
+        );
+      } else if (c) {
```
kozlovic, line 102: Nit: if we are in a `else` of `if (c === null)` what is the purpose of `if (c)`? would a simple `else` be enough? If not, what other value can `c` have and what are we doing with that condition?

README.md
aricart: the timeout is there as an option, this way we can have other options like headers - the timeout is set by default to 1000, whether it is provided or not.

nats-base-client/types.ts
aricart: sorting them - this is outstanding, so I moved it so I could track it - working in this feature atm.

src/deno_transport.ts
aricart: The check for null is special, because it API call can return number or null. Null means EOF. with the check for (c), if it is greater than zero, we process otherwise we ignore.

README.md
aricart: Also thinking of moving the message payload as part of the option for publish/request. The reason being is that the default arguments work well until you need to specify one of them. For request, the only required argument is the subject, the rest are defaulted, same for publish. This simplifies the syntax, as now you can specify an optional value, without having to put a dummy value on the previous optional argument:

`publish/request('foo', {headers: h})` vs. `publish/request('foo', '', {headers: h})` in the case where the payload is not a string, it becomes more problematic. In the case of the request, the timeout will always be set to a default of 1000, unless client provides a different value.

kozlovic: Yes, but if payload is optional, what will be the syntax when user wants to provide it? Does that becomes `publish('foo', {payload: 'bar', headers: h})`? So in more common situations (presence of a payload), there is more to "type" if you will.

aricart: yeap.

PR #13, opened 2020-07-31
[feat] added support for stale-connection checks (maxPingOut, and pingInterval)

aricart:
Added support for stale connection checks.


nats-base-client/heartbeats.ts
```diff
@@ -0,0 +1,90 @@
...
+  pendings: Promise<void>[] = [];
+
+  constructor(ph: PH, interval: number, maxOut: number) {
+    this.ph = ph;
+    this.interval = interval;
+    this.maxOut = maxOut;
+  }
+
+  // api to start the heartbeats, since this can be
+  // spuriously called from dial, insure we don't
```
kozlovic, line 38: `ensure`

nats-base-client/heartbeats.ts
```diff
@@ -0,0 +1,90 @@
...
+  // leak timers
+  start() {
+    this.cancel();
+    this._schedule();
+  }
+
+  // api for canceling the heartbeats, if a string is
```
kozlovic, line 45: I don't understand `if a string is provided..` where?

aricart: stale comment.

nats-base-client/heartbeats.ts
aricart: fixed

nats-base-client/heartbeats.ts
aricart: fixed

PR #21, opened 2020-08-06
[chore] simplified connection options by removing URL as really we just handle hostports

aricart:
- [BREAKING] removed `url` connection option. Instead, provide `servers` - this option can be a string or an array of strings.

### DIFF
--- a/workflows/test.yml
+++ b/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Set NATS Server Version
         run: echo '::set-env name=NATS_VERSION::v2.1.7'
-      - name: Get nats-server
+      - name: Get nats-servers
         run: |
           wget "https://github.com/nats-io/nats-server/releases/download/$NATS_VERSION/nats-server-$NATS_VERSION-linux-amd64.zip" -O tmp.zip
           unzip tmp.zip
--- a/README.md
+++ b/README.md
@@ -54,8 +54,8 @@ import { connect, NatsConnection } from "src/mod.ts";
 // common options that you could pass look like:
 const localhostAtStandardPort = {};
 const localPort = { port: 4222 };
-const hostAtStdPort = { url: "demo.nats.io" };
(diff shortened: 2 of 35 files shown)


workflows/test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
 
       - name: Set NATS Server Version
         run: echo '::set-env name=NATS_VERSION::v2.1.7'
-      - name: Get nats-server
+      - name: Get nats-servers
```
kozlovic, line 34: Is that a rename-all kind of things? Not sure why you would have change nats-server to nats-servers here :-)

aricart: weird my local version doesn't have it. - I will fix the title on merge.

haywirez:
Upgrading an older brownfield project and running into connection issues — do I read this correctly, support for `nats://lollipop:SOME_PASSWORD@dokku-nats-lollipop:4222/lollipop` style connection URLs has been dropped in favor of hostPorts and credentials provided separately? Is there a parsing utility function available? Asking as projects with plugins such as https://github.com/dokku/dokku-nats still do connection URLs as the default (paging @josegonzalez)

josegonzalez:
Providing a DSN is probably the most standard way to provide connection information and the Dokku project will not be offering any official alternative for plugins it maintains. I would recommend wrapping `nats.deno` with something that parses the DSN and sets up the connection internally if the official library won't support it.

haywirez:
In case someone finds this issue, a naive 8 liner:

```ts
const parseNatsUrl = (url) => {
  if (url.indexOf('nats://') === -1) {
    return { servers: url }
  }
  const [usernamePassword, hostPort] = url.replace('nats://', '').split('@')
  const [user, pass] = usernamePassword.split(':')
  return { user, pass, servers: hostPort }
}
```

It's probably wiser to use something like [@soluble/dsn-parser](https://www.npmjs.com/package/@soluble/dsn-parser)

PR #57, opened 2020-09-04
client ip is accessible from the `info` property in the connection

aricart:


nats-base-client/types.ts
```diff
@@ -136,18 +137,28 @@ export interface Base {
...
-  version: string;
-  echo?: boolean;
+  jetstream?: boolean;
+  client_id?: number;
+  client_ip?: string;
   nonce?: string;
-  nkey?: string;
+  cluster?: string;
+  connect_urls?: string[];
+  ws_connect_urls?: string[];
```
kozlovic, line 160: This is used only between servers. The client will get it from `connect_urls`. For WS clients, it will be WS urls. In other words, the server always send `connect_urls` but adapt the content based on the type of client.

nats-base-client/types.ts
```diff
@@ -136,18 +137,28 @@ export interface Base {
...
+  host: string;
+  port: number;
+  headers?: boolean;
+  auth_required?: boolean;
   tls_required?: boolean;
   tls_verify?: boolean;
-  connect_urls?: string[];
+  tls_available?: boolean;
   max_payload: number;
-  client_id: number;
```
kozlovic, line 143: I already commented in the past that those constant re-ordering are making review difficult because I really have to scan to see what is added/removed vs just moved around.

kozlovic (changes requested):

nats-base-client/types.ts
aricart: yes well - I needed to map them to the server.

nats-base-client/types.ts
aricart: awesome thanks.

nats-base-client/types.ts
kozlovic: What I mean is that some move around and it is difficult to see which ones are added or removed, or simply moved. Why move the ones that did not change?

aricart: They now are ordered in teh same way as the server. - I prefer alpha, but server is not sorted.

kozlovic: You pick the order you want, they don't have to align with what is in the server's Info structure, but then please stick with it.

kozlovic (approved):
LGTM

PR #58, opened 2020-09-11
[feat] respond allows specifying a reply subject

aricart:
JetStream requires some of the respond iterations to include a reply subject (+NXT)


kozlovic (approved):
LGTM

PR #64, opened 2020-09-22
websocket client urls/updates

aricart:
By default websocket clients will use the `wss://` protocol and connect to a NATS server. In cases where they are connecting to a proxy, it becomes possible to not only provide the `host:port` portion, but the requirement may be a full URL with a path. The proxy may have some semantic associated with the path portion.

A `URLParseFn` type enables a Transport to pre-process all URLs provided during connect - this allows re-targeting default ports to match for example default `ws://` or `wss://` ports, before the underlying hostport parsing activates.

To capture the initial values specified by a client, such as `ws://server:port/path`, the initial hp specification as modified by `URLParseFn` or not is stored in the `src` property. 

Finally, in cases where they are connecting through a proxy, added option to `ignoreClusterUpdates` to enable the client to specify to ignore them, this can be turned off in the server as well.

### DIFF
--- a/nats-base-client/protocol.ts
+++ b/nats-base-client/protocol.ts
@@ -37,7 +37,7 @@ import {
 } from "./util.ts";
 import { nuid } from "./nuid.ts";
 import { DataBuffer } from "./databuffer.ts";
-import { Server, Servers } from "./servers.ts";
+import { ServerImpl, Servers } from "./servers.ts";
 import { Dispatcher, QueuedIterator } from "./queued_iterator.ts";
 import type { MsgHdrs, MsgHdrsImpl } from "./headers.ts";
 import { SubscriptionImpl } from "./subscription.ts";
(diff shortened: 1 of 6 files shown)


kozlovic (approved):
LGTM, but not related directly to this PR, but not sure you needed the "firstServer" option. Since you have the option `randomize`, if false, it would take the first server from the array and so not sure why you explicitly need to have "firstServer" option.

aricart:
> LGTM, but not related directly to this PR, but not sure you needed the "firstServer" option. Since you have the option `randomize`, if false, it would take the first server from the array and so not sure why you explicitly need to have "firstServer" option.

`firstServer` I think is scheduled for extinction. The original code had `url` and `servers`, and the first server would have been akin to `{url: x, servers: [y, z]}`  - let me do some digging on the options, but likely that option just needs to go away.

kozlovic (approved):
LGTM

PR #70, opened 2020-09-29
connect sends extraneous options

aricart:
the `Connect` object was extended with all connection options rather than just the ones expected by the server,


kozlovic (approved):
LGTM

PR #74, opened 2020-10-16
mark info received right when the first info is processed

### LINKED ISSUE #73: connect protocol can be sent multiple times on connect
Due to timing issues with resolving the connection, it is possible for the CONNECT proto to be sent twice if it receives a second INFO before the connection resolves. This results in a disconnect from the server

aricart:
The receiving of the info was noted when the PONG for the connection was resolved. If the server sent another info within this timeframe, a double connect is sent resulting in the server disconnecting the client.

FIX #73


kozlovic (approved):
LGTM!

PR #76, opened 2020-10-16
added ability to remember the tlsHost name

### LINKED ISSUE #75: tls certificate problem with gossiped servers
Gossiped servers are presented as host:port - if the original name used for connecting is not saved, certificate validation fails during tlsUpgrade on Deno platform

aricart:
FIX #75


kozlovic (approved):
LGTM.. although I wonder if we could have just made it an option that user would have to set to avoid adding IP parsing code. Of course, user experience is better if they don't have to provide an option and library can figure out everything on its own.

aricart:
I agree for client code, but didn't want to introduce extra complexity to all the tools downstream, which then would have to telegraph that option.

PR #80, opened 2020-11-11
The status can now have a description.

### LINKED ISSUE #79: headers can now have status/description
The header status line can now have status/description. Status is now stored in `Status` and description in `Description`.

aricart:
Added internal property `description`.
Changed internal `error` property to be `code`.
Added computed property `hasError` to return true if code is not 2xx.
Added computed property `status` that returns a formatted string with code/status.

FIX #79

### DIFF
--- a/nats-base-client/headers.ts
+++ b/nats-base-client/headers.ts
@@ -19,6 +19,8 @@ import { ErrorCode, NatsError } from "./error.ts";
 import { TD, TE } from "./encoders.ts";
 
 export interface MsgHdrs extends Iterable<[string, string[]]> {
+  hasError: boolean;
+  status: string;
   get(k: string): string;
   set(k: string, v: string): void;
   append(k: string, v: string): void;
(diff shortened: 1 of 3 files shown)


nats-base-client/headers.ts
```diff
@@ -81,8 +85,13 @@ export class MsgHdrsImpl implements MsgHdrs {
     const lines = s.split(MsgHdrsImpl.CRLF);
     const h = lines[0];
     if (h !== MsgHdrsImpl.HEADER) {
-      const str = h.replace(MsgHdrsImpl.HEADER, "");
-      mh.error = parseInt(str, 10);
+      let str = h.replace(MsgHdrsImpl.HEADER, "");
+      mh.code = parseInt(str, 10);
+      const scode = mh.status.toString();
+      mh.set("Status", scode);
+      str = str.replace(mh.status.toString(), "");
```
kozlovic, line 92: would `mh.status.toString()` not be `scode`?

nats-base-client/headers.ts
```diff
@@ -81,8 +85,13 @@ export class MsgHdrsImpl implements MsgHdrs {
...
+      mh.description = str.trim();
+      mh.set("Description", mh.description);
```
kozlovic, line 94: I would add "Description" header only if `mh.description` is not empty...

nats-base-client/headers.ts
aricart: the code above is completely broken - some undo killed the value.

kozlovic (approved):
LGTM

PR #82, opened 2020-11-16
fix publisher connection drain

aricart:
[fix] publisher connection drain didn't flush pending buffer after draining connections
[fix] tests were incorrectly flushing masking issue


kozlovic (approved):
LGTM

PR #84, opened 2020-12-08
[feat] added `reply` option to RequestOptions 

aricart:
This change makes it easy to `request()` while specifying a reply subject that is in the client's subscribe permission set.


kozlovic (approved):
LGTM

PR #88, opened 2021-01-06
deno bundler safari compatibility

aricart:
[fix] removed class field initializers and static fields as the current deno bundler doesn't polyfill.


kozlovic (approved):
LGTM

PR #91, opened 2021-01-25
connection timeout can fire in an 'unhandled' situation

aricart:
The timer creation happens outside of the try catches, so if the initial transport connect fails, it is possible for the timer to trigger in an unhandled context.

See https://github.com/nats-io/nats.js/issues/376


kozlovic (approved):
LGTM

PR #98, opened 2021-02-03
process error enhancements and fix

aricart:
- [NEW] Added support for authentication expired.
- [NEW] Added Events.Error to provide a mechanism to listen for errors
- [CHANGED] 2 auth errors in a row now correctly closes the client
- [FIX] Process error was calling close, but in the context of connecting if the connection wasn't resolved, it would continue to try to connect after the close.


kozlovic (approved):
LGTM

PR #99, opened 2021-02-07
removed client options `headers` and `noResponders` - features auto-set if server supports headers.

aricart:
javascript clients were exposing `headers` and `noResponder` optionsto enable/disable these features.
Go client simply keys from server support. Changed to match.


matthiashanel (approved):
LGTM

PR #103, opened 2021-02-10
support an inboxPrefix in the options of connect()

brimworks:
This is needed so that JWT tokens can enforce per-user _INBOX prefixes which can only be subscribed to by the individual user, but all other users can publish to that per-user _INBOX prefix. Note that the Java client also supports this option:

https://javadoc.io/doc/io.nats/jnats/latest/io/nats/client/Options.Builder.html#inboxPrefix-java.lang.String-

### DIFF
--- a/nats-base-client/protocol.ts
+++ b/nats-base-client/protocol.ts
@@ -54,8 +54,13 @@ const FLUSH_THRESHOLD = 1024 * 32;
 
 export const INFO = /^INFO\s+([^\r\n]+)\r\n/i;
 
-export function createInbox(): string {
-  return `_INBOX.${nuid.next()}`;
+export function createInbox(inboxPrefix?: string): string {
+  if (typeof inboxPrefix === "string") {
+    inboxPrefix = inboxPrefix + ".";
+  } else {
+    inboxPrefix = "";
+  }
+  return `${inboxPrefix}_INBOX.${nuid.next()}`;
 }
 
 const PONG_CMD = fastEncoder("PONG\r\n");
@@ -684,7 +689,7 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
   private initMux(): void {
     const mux = this.subscriptions.getMux();
     if (!mux) {
-      const inbox = this.muxSubscriptions.init();
+      const inbox = this.muxSubscriptions.init(this.options.inboxPrefix);
       // dot is already part of mux
       const sub = new SubscriptionImpl(this, `${inbox}*`);
       sub.callback = this.muxSubscriptions.dispatcher();
--- a/nats-base-client/protocol.ts
+++ b/nats-base-client/protocol.ts
@@ -15,6 +15,7 @@
 import {
   ConnectionOptions,
   DebugEvents,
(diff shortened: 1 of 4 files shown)


nats-base-client/protocol.ts
```diff
@@ -54,8 +54,13 @@ const FLUSH_THRESHOLD = 1024 * 32;
...
 
-export function createInbox(): string {
-  return `_INBOX.${nuid.next()}`;
+export function createInbox(inboxPrefix?: string): string {
+  if (typeof inboxPrefix === "string") {
+    inboxPrefix = inboxPrefix + ".";
+  } else {
+    inboxPrefix = "";
+  }
+  return `${inboxPrefix}_INBOX.${nuid.next()}`;
```
aricart, line 63: I would make the default prefix be _INBOX prefix, which can then be replaced by the user-specified value.

aricart (changes requested):
rather than make _INBOX part of the prefix, I would just make _INBOX the default prefix, to be replaced by the user specified option. This means that the handling is simpler since we just have a constant or the replaced value and the format just uses the initialized prefix.

aricart:
@brimworks If you want me to edit your PR I can do that as well

brimworks:
Sorry, I was on vacation this last week. I'll make that change, or if you want to edit the PR that is fine too.

aricart:
I'll do it today.

brimworks:
I went ahead and made the suggested change here: https://github.com/nats-io/nats.deno/pull/103/commits/79ec1436c9989a85bbd9c5e7c95737070a151c76

I also assume that I should follow the `nats.java` client lead and have the default `inboxPrefix` include the trailing dot: https://github.com/nats-io/nats.java/blob/master/src/main/java/io/nats/client/Options.java#L187

brimworks:
I'll also go ahead and merge in the last code to resolve the conflict.

aricart:
Actually - this doesn't work right - `createInbox` is an imported function (it has no association with the connection which specifies the option). This means that there's a race condition between connections. The last connection to set it will create incorrectly named inboxes for the other clients.

aricart:
This actually is more subtle than expected - I will provide the implementation of this before release. Thanks for bringing it to my attention.

brimworks:
> Actually - this doesn't work right - createInbox is an imported function (it has no association with the connection which specifies the option). This means that there's a race condition between connections. The last connection to set it will create incorrectly named inboxes for the other clients.

I don't follow what you mean. I changed `createInbox()` function so that it takes an optional inbox prefix, so there should be no race condition regarding the inbox prefix. Maybe there is some other unrelated race condition?

aricart:
yes internally the uses of createInbox would have to telegraph all the way out - the fix is somewhat easy, but some of the tests fail afterward.

aricart:
@brimworks I reworked the PR to cover other areas where createInbox is used internally.
https://github.com/nats-io/nats.deno/pull/108

aricart:
@brimworks thanks for the PR and bringing the need for this feature.

PR #105, opened 2021-02-16
Preparing for JetStream

aricart:
- [feat] added an internal yield callback on the subscription to enable support of auto-acks on JetStream
- [feat] added an internal cleanup function for when a subscription is called (such as removing a consumer)
- [doc] deno now formats markdown :)


kozlovic (approved):
LGTM

PR #106, opened 2021-02-16
FIX nats-io/nats.ws/issues/80

aricart:
- [change] refactored transport factory to also provide `defaultPort`, `urlParseFn`, and `factory` - `defaultPort` and `urlParseFn` are transport specific (ie websockets)

- [change] made default hostport calculated from the transport information.
- [removed] default hostport constant
- [fix] lint warnings for snake case json fields

FIXES https://github.com/nats-io/nats.ws/issues/80


kozlovic (approved):
LGTM

PR #107, opened 2021-02-17
[feat] allow non-file tls certificate configuration

aricart:
- [feat] added `tls.cert`, `tls.key`, `tls.ca` to make it possible to specify certificate as a string on clients that would allow that (node).
- FIX https://github.com/nats-io/nats.js/issues/395


kozlovic (approved):
LGTM

PR #108, opened 2021-02-18
added `inboxPrefix` as a connection option

aricart:
- [feat] added `inboxPrefix` option to set the prefix of subjects generated by createInbox.
- [change] modified `createInbox` to take an optional argument to override `_INBOX` as the prefix.


kozlovic (approved):
LGTM

nats-base-client/protocol.ts
```diff
@@ -54,8 +54,12 @@ const FLUSH_THRESHOLD = 1024 * 32;
...
 export const INFO = /^INFO\s+([^\r\n]+)\r\n/i;
 
-export function createInbox(): string {
-  return `_INBOX.${nuid.next()}`;
+export function createInbox(prefix: string = ""): string {
+  prefix = prefix || "_INBOX";
+  if (typeof prefix !== "string") {
+    throw (new Error("prefix must be a string"));
+  }
+  return `${prefix}.${nuid.next()}`;
```
brimworks, line 62: Don't we want to have the default prefix to also have a trailing dot so we can match how the Java client works? Also, the Java client allows one to set the inbox prefix to the empty string.

https://github.com/nats-io/nats.java/blob/master/src/main/java/io/nats/client/Options.java#L187
https://github.com/nats-io/nats.java/blob/master/src/main/java/io/nats/client/impl/NatsConnection.java#L942

aricart:
The trailing dot is only used for mux. Create inbox should return invalid
subjects.

PR #111, opened 2021-02-25
fix untyped codec implementation

aricart:
- [fix] missing type from the JSON codec which the compiler was angry about
- [fix] sample had rogue `boolean` option
- [doc] added documentation on the JSONCodec<T> feature


kozlovic (approved):
LGTM

PR #112, opened 2021-02-27
[change] [breaking] event names now follow ES5 naming

aricart:
chore to cleanup event names to be ES5 compliant


matthiashanel (approved):
LGTM

PR #113, opened 2021-02-27
support downstream sub closed

aricart:
[change] added an internal promise that resolves when a subscription is closed - this will benefit downstream wrappers such as jetstream to detect when their underlying subscription is closed.


kozlovic (approved):
LGTM

PR #114, opened 2021-03-03
simplify header code checks downstream

aricart:
- [fix] changed header code interface to be always a number simplifying downstream logic
- [fix] linter warning


kozlovic (approved):
LGTM

PR #120, opened 2021-03-13
[fix] leak of implementation type to JS api (QueuedIterator is now an interface

aricart:
this fixes implementation details to clients


kozlovic (approved):
LGTM

PR #122, opened 2021-03-13
normalize error handling

aricart:
- [fix] normalized error code handling:
    - codes are string representation of the number
    - description is lowercased string
    - constants simply reference the string code


kozlovic (approved):
LGTM

PR #123, opened 2021-03-14
Subscribe requires deliverto

aricart:
[fix] subscribe requires `deliverTo` or it is rejected, likewise pull subscribers cannot specify `deliverTo`
[fail] marked test cross-account pull consumer as failing - needs investigation, but the client is sending a request to a subject that the server can respond to


kozlovic (approved):
LGTM

PR #124, opened 2021-03-14
basic js samples

aricart:
[doc] some simple examples of using jetstream
[fix] fixed some of the function signatures to be partials to allow inline of partial configurations.
[fix] exported missing JS types


examples/jetstream/02_push_consumer.ts
```diff
@@ -1,23 +1,34 @@
...
+const subj = nuid.next();
+await jsm.streams.add({ name: stream, subjects: [`${subj}.>`] });
+
+// create a consumer that delivers to the subscription
+await jsm.consumers.add(stream, {
   ack_policy: AckPolicy.Explicit,
   deliver_subject: "my.messages",
 });
 
+// publish some old nats messages
```
ripienaar, line 26: `some old nats messages` is weird..

examples/jetstream/04_ephemeral_consumer.ts
```diff
@@ -1,18 +1,18 @@
...
-  manualAcks: false, // auto ack messages
-  max: 3,
-});
-const done = (async () => {
+const js = nc.jetstream();
+const opts = consumerOpts();
+opts.manualAck();
+opts.maxMessages(2);
+opts.deliverTo("xxx");
+const sub = await js.subscribe("a.>", opts);
```
ripienaar, line 10: shouldnt this subscribe match deliverto?

ripienaar (approved):
seems good, dont know TS, just some comments there

examples/jetstream/04_ephemeral_consumer.ts
aricart: In this case the subscribe creates a consumer if not there, so it can only find the stream based on a subject mapped to the stream.

ripienaar:
LGTM

Issue #128, opened 2021-03-17
Lister should be an AsyncIterator


nleguen:
jsm.streams.list() or jsm.consumers.list(stream) returns a Lister. Any reason why it cannot be used as an AsyncIterator ?

aricart:
List is a paged result, `next()` is returning a list for the page. Not sure how that is helpful outside of the context of retrieving ALL streams and ALL consumers. But willing to consider it.

nleguen:
#129

PR #129, opened 2021-03-18
implements AsyncIterable<T> for Lister<T>

nleguen:
When needed to retrieve all streams or consumers instead of a paged list, this simple change makes possible to do something like:
```javascript
for await(const info of jsm.streams.list()) {
  console.log(info.config.name); // do whatever you want with the info
}
```
It doesn't break the actual API: `jsm.streams.list().next()` still returns a list for the paged result.


aricart (approved):
LGTM

aricart:
@nleguen thanks for the contribution

aricart:
FIX #128

PR #134, opened 2021-03-26
fix incorrect copy of reply subject

aricart:
[fix] the reply subject buffer was copied from the offset of the reply rather than the subject, providing the client with an incorrect reply subject.


kozlovic (approved):
LGTM

PR #141, opened 2021-04-09
remove use of nullish operator as react-create-app will fail to build

aricart:
Replaced all uses of the nullish operator `??` with `||` as downstream react applications created by `react-create-app` will fail.


matthiashanel (approved):
LGTM

PR #142, opened 2021-04-12
support subscription timeout

aricart:
- [FIX] fix callback subscriptions to report the timeout error, added tests to verify subscription, noMux requests, and that JetStream pull/fetch requests also trigger a timeout.
- [FEAT] subscriptions now expose `setTimeout(millis)` and `cancelTimeout()`. The initial optional subscription timeout option is cleared on first message received. With the expose API it becomes possible for a subscription to re-set a timer for the next message. This can be useful when implementing subscriptions that expect certain number of messages within some amount of time.
- [FIX] changed error message expected from the server on a test
- [FIX] remove n/a comment


wallyqs (approved):
LGTM

nats-base-client/protocol.ts
```diff
@@ -378,6 +378,7 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
     sub.received += 1;
 
     if (sub.callback) {
+      sub.cancelTimeout();
```
kozlovic, line 381: Not sure what semantics you want for a subscription timeout, but if you cancel here before invoking callback, basically saying we got a message, so stop the timeout timer, should you not set it again after the callback? 
In other words, the way I read it, you set the timeout in the constructor and timeout is stopped when a message is received. But the subscription will not timeout again, even if no message is received after that?

aricart:
They redo the timeout if they want it. The auto angel is the only thing we
do.

aricart:
Withdrawing this.

PR #145, opened 2021-04-15
[feat] added ability to opt-out from mime formatted header names

aricart:
added an optional argument that allows an application's use of headers to opt-out of  mime header formatting. To retrieve values, they must also opt-out


nats-base-client/headers.ts
```diff
@@ -182,24 +182,24 @@ export class MsgHdrsImpl implements MsgHdrs {
     return k.trim();
   }
 
-  get(k: string): string {
-    const key = MsgHdrsImpl.canonicalMIMEHeaderKey(k);
+  get(k: string, mime = true): string {
+    const key = mime ? MsgHdrsImpl.canonicalMIMEHeaderKey(k) : k;
```
wallyqs, line 186: The check would work better being case insensitive, during the transport of the headers any library may transform the header etc.. but retrieving a key "foo", should be the same as getting "Foo" for example, but then also may want an option to get the exact match.

nats-base-client/headers.ts
```diff
@@ -182,24 +182,24 @@ export class MsgHdrsImpl implements MsgHdrs {
...
-  has(k: string): boolean {
-    return this.get(k) !== "";
+  has(k: string, mime = true): boolean {
+    return this.get(k, mime) !== "";
   }
 
-  set(k: string, v: string): void {
-    const key = MsgHdrsImpl.canonicalMIMEHeaderKey(k);
+  set(k: string, v: string, mime = true): void {
+    const key = mime ? MsgHdrsImpl.canonicalMIMEHeaderKey(k) : k;
```
wallyqs, line 196: I would avoid transforming the header key here and just take it as is, then only when doing a get have an option of whether the library is expecting the case to preserved or not.

aricart:
Implemented as specified here: https://github.com/nats-io/nats.deno/issues/147

aricart:
nats.deno is going to implement it as described in https://github.com/nats-io/nats.deno/issues/147.

- set() will write header as per the user
- get() will return the first value in the first key that matches in a case insensitive way
- values() return the union of all keys that match case-insensitively.  

No key normalization, headers are insensitive, if the user wants to store a canonical header, they can use the provided function to transform the key.

wallyqs (approved):
LGTM

aricart:
After speaking with Colin:

All options with the exception of append, have an `exact` option which will constrain any operation to be case-sensitive 
only. This optional option (false by default) will:
- `set()`, only replaces an exact match header
- `get()`, only returns an exact match header
- `values()` only returns an exact match header

`append` doesn't provide this setting, as any append is case-preserving, so only adds to an exact match.

nats-base-client/headers.ts
```diff
@@ -182,41 +180,74 @@ export class MsgHdrsImpl implements MsgHdrs {
...
+  has(k: string, exact = false): boolean {
+    return this.findKeys(k, exact).length > 0;
+  }
+
+  set(k: string, v: string, exact = false): void {
+    this.delete(k, exact);
+    this.append(k, v);
   }
 
   append(k: string, v: string): void {
```
kozlovic, line 222: Since you have an `exact` option in other APIs, should you have one here too?

nats-base-client/headers.ts
```diff
@@ -182,41 +180,74 @@ export class MsgHdrsImpl implements MsgHdrs {
...
-    const key = MsgHdrsImpl.canonicalMIMEHeaderKey(k);
+    // validate the key
+    MsgHdrsImpl.canonicalMIMEHeaderKey(k);
```
kozlovic, line 224: What does this do? Your comment says "validate", but does that change it? And valid based on what, the case or special characters, etc??

nats-base-client/headers.ts
```diff
@@ -182,41 +180,74 @@ export class MsgHdrsImpl implements MsgHdrs {
...
     const value = MsgHdrsImpl.validHeaderValue(v);
-    let a = this.headers.get(key);
+    let a = this.headers.get(k);
```
kozlovic, line 226: If you had the `exact` option, you would propagate it to the get() and set() below.

nats-base-client/headers.ts
aricart: the key is validated, the validation will catch any key that contains characters that are not valid in headers, as per the header specifications (non-printable, etc)

nats-base-client/headers.ts
aricart: changed see new

nats-base-client/headers.ts
aricart: append always matches, but this changed in the current

nats-base-client/headers.ts
aricart: this is old

nats-base-client/headers.ts
```diff
@@ -182,41 +185,89 @@ export class MsgHdrsImpl implements MsgHdrs {
...
+    return "";
+  }
+
+  has(k: string, match = Match.Exact): boolean {
+    return this.findKeys(k, match).length > 0;
+  }
+
+  set(k: string, v: string, match = Match.Exact): void {
+    this.delete(k, match);
+    if (match === Match.CanonicalMIME) {
```
kozlovic, line 233: Since you rely on `append()`, which does that check, why do you do it here?

kozlovic (approved):
LGTM, minor comment about performing canonicalMIMEHeaderKey() twice on a set?

nats-base-client/headers.ts
aricart: Thank you good catch!

kozlovic (approved):
LGTM

aricart:
@kozlovic raised the point that append what preserving the input's case, but options provided could be in disagreement.

kozlovic (approved):
LGTM

Issue #144, opened 2021-04-15
Issue with Deno 1.9.0


edward653:
Upgraded to Deno 1.9.0 and alongside this, went from nats 1.0.0_rc4 to 1.0.2. Build fails with the following error:

error: TS2345 [ERROR]: Argument of type 'Conn<Addr>' is not assignable to parameter of type 'Conn<NetAddr>'.
  Type 'Addr' is not assignable to type 'NetAddr'.
    Type 'UnixAddr' is missing the following properties from type 'NetAddr': hostname, port
      this.conn,
      ~~~~~~~~~
    at https://deno.land/x/nats@v1.0.2/src/deno_transport.ts:144:7


I've not yet managed to dig into the cause, as yet. This doesn't seem to stem from my usage of the module.

keawade:
I've been seeing this issue as well trying to use the demo code.

Code to reproduce:

```ts
// import the connect function
import { connect } from "https://deno.land/x/nats@v1.0.2/src/mod.ts";

const servers = [
  {},
  { servers: ["demo.nats.io:4442", "demo.nats.io:4222"] },
  { servers: "demo.nats.io:4443" },
  { port: 4222 },
  { servers: "localhost" },
];
await servers.forEach(async (v) => {
  try {
    const nc = await connect(v);
    console.log(`connected to ${nc.getServer()}`);
    // this promise indicates the client closed
    const done = nc.closed();
    // do something with the connection

    // close the connection
    await nc.close();
    // check if the close was OK
    const err = await done;
    if (err) {
      console.log(`error closing:`, err);
    }
  } catch (err) {
    console.log(`error connecting to ${JSON.stringify(v)}`);
  }
});
```

```shell
~/Projects/deno-nats-test$ deno run --unstable --allow-net mod.ts
Check file:///home/keawade/Projects/deno-nats-test/mod.ts
error: TS2345 [ERROR]: Argument of type 'Conn<Addr>' is not assignable to parameter of type 'Conn<NetAddr>'.
  Type 'Addr' is not assignable to type 'NetAddr'.
    Type 'UnixAddr' is missing the following properties from type 'NetAddr': hostname, port
      this.conn,
      ~~~~~~~~~
    at https://deno.land/x/nats@v1.0.2/src/deno_transport.ts:144:7
~/Projects/deno-nats-test$ deno --version
deno 1.9.0 (release, x86_64-unknown-linux-gnu)
v8 9.1.269.5
typescript 4.2.2
```

[...]

PR #146, opened 2021-04-16
update Deno dependencies for 1.9.0

aricart:
- [updated] to be compatible with deno 1.9.0, 1.9.0 breaks compatibility on connection types, as `Conn` is now a generic type. The current fix adds compilation ignore which allows pre-1.9.0 clients to ignore the generic version of `Conn`, and allows 1.9.0 to be happy.
- [updated] std lib dependencies to 0.90.0
- [fix] linter issues exposed by 1.9.0


aricart:
This also fixes https://github.com/nats-io/nats.deno/issues/144

matthiashanel (approved):
LGTM

PR #151, opened 2021-04-23
infra to build js-cluster for tests

aricart:


tests/helpers/cluster.ts
```diff
@@ -28,27 +28,36 @@ const argv = parse(
...
+    : await NatsServer.cluster(
+      argv.count,
+      { port: argv.port },
+      argv.debug,
+    );
+
   cluster.forEach((s) => {
     console.log(
-      `launched server [${s.process.pid}] at ${s.hostname}:${s.port} - monitor ${s.monitoring}`,
+      `nats-server -c ${s.configFile} [${s.process.pid}] at nats://${s.hostname}:${s.port} cluster://${s.hostname}:${s.cluster} http://127.0.0.1:${s.monitoring}`,
```
kozlovic, line 60: With JS, for clustering to work well, the routes need to be explicit because the replicas and quorum size is based on that. Meaning that a server should have 
```
cluster {
   port: xxx
   routes [
       "nats://thisserver:routeport",
       "nats://2ndserver:routeport",
       "nats://3rdserver:routeport",
   ]
}
```
So I am not sure if having a server just with cluster port running is enough.

aricart: They are:
All the cluster ports are captured here:

https://github.com/nats-io/nats.deno/pull/151/files#diff-13f7841d0f019c3a5a1be00d61f041840c0254418db41c57810cefbc11354763R290

And then after all captured, self-routes are removed (I know they don't have to be)

https://github.com/nats-io/nats.deno/pull/151/files#diff-13f7841d0f019c3a5a1be00d61f041840c0254418db41c57810cefbc11354763R305

And finally, all the servers are restarted with their JS cluster configurations.

aricart: A full config looks like this:
```
port: 4222
cluster {
  name: MOYI2Y0O2HYD057TSX017D
  listen: "127.0.0.1:49922"
  routes [
    nats://127.0.0.1:49928
   ]
 }
ports_file_dir: /var/folders/1y/_rk6yz81341g5q8k1vr5nqmw0000gq/T
host: "127.0.0.1"
http: "127.0.0.1:49921"
jetstream {
  max_file_store: 1048576
  max_memory_store: 1048576
  store_dir: /tmp/jetstream/MOYI2Y0O2HYD057TSX01ED
 }
server_name: MOYI2Y0O2HYD057TSX01AV
```

kozlovic (approved):
LGTM

PR #155, opened 2021-05-13
[fix] fixed an issue where preventing double-acks prevented repeat invocations of `working`

### LINKED ISSUE #154: JS: WPI (jsmsg.working()) is not resending as first ack, will prevent others.
the jsmsg ack added a regression to prevent double acks (ie, user handler nack, but then an auto-ack was set). This restriction is only necessary if a non-terminal ack is sent (ack, nak, term).

aricart:
FIX #154


kozlovic (approved):
LGTM

PR #156, opened 2021-05-13
[fix/feat] consumer options builder doesn't provide access to all the options possible

### LINKED ISSUE #153: jetstream consumer builder
`consumerOpts()` doesn't allow passing some complex options to create a consumer - Instead of telegraphing all options, it should allow passing a Partial<ConsumerConfig>

aricart:
Added the ability to initialize a builder with the consumer options, this allows the consumer to be further customized.

FIX #153


kozlovic (approved):
LGTM

PR #162, opened 2021-06-16
exposed and enabled queue option for consumer opts

aricart:
[FIX] js.subscribe followed nats.go, and option for queue was not exposed, it seems this is now supported.

FIX #nats-io/nats.js/issues/432


kozlovic (approved):
LGTM

PR #164, opened 2021-06-29
Added ConsumerConfig idle_heartbeat and flow_control options

aricart:
[FEAT] options `idle_heartbeat` and `flow_control` to `ConsumerConfig`
[FEAT] added `idleHeartbeat(millis)` and `flowControl()` to `ConsumerOptsBuilder`
[FEAT] added `isHeartbeatMsg()` to simplify checking if a raw message sent to a NATS subscription is a JetStream heartbeat message.
[FEAT] added examples for idle_heartbeat and flow_control
[FEAT] exposed `closed` property on subscription objects - this is a promise that is resolved when a subscription is finished.

### DIFF
--- a/examples/jetstream/09_flowcontrol.ts
+++ b/examples/jetstream/09_flowcontrol.ts
@@ -1,0 +1,48 @@
+import { AckPolicy, connect, toJsMsg } from "../../src/mod.ts";
+import { nuid } from "../../nats-base-client/nuid.ts";
+import { isFlowControlMsg } from "../../nats-base-client/jsutil.ts";
+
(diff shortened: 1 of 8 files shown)


kozlovic (commented):
I think those protocols need to be handled by the library, not have the user do the work. The go client for instance does that (plugged into processing of a message from the connection and detect as a control message (fc/hearbeat) and don't pass it to the subscription callback)

aricart:
Yes it does that unless you create a plain nats subscription

kozlovic (approved):
LGTM

PR #165, opened 2021-06-29
added `max_msgs_per_subject` to `StreamConfig`

aricart:
[FEAT] added `max_msgs_per_subject` option to `StreamConfig`, this option specifies on wildcard streams how many messages to keep per subject.


kozlovic (approved):
LGTM

PR #166, opened 2021-06-29
last_by_subject JSM.streams.getMessage()

aricart:
[FEAT] [BREAKING] Slight change to JSM streams `getMessage()` the API the second argument changed from a sequence number to be either `{seq: #}` or `last_by_subject: "my.subject"`, this enables the feature for retrieving the last message in a stream with the specified subject.

The JetStream features in the JavaScript clients are pre-release and subject to changes.

### DIFF
--- a/nats-base-client/types.ts
+++ b/nats-base-client/types.ts
@@ -389,7 +389,7 @@ export interface StreamAPI {
   delete(stream: string): Promise<boolean>;
   list(): Lister<StreamInfo>;
   deleteMessage(stream: string, seq: number, erase?: boolean): Promise<boolean>;
-  getMessage(stream: string, seq: number): Promise<StoredMsg>;
+  getMessage(stream: string, query: MsgRequest): Promise<StoredMsg>;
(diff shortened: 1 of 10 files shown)


nats-base-client/types.ts
```diff
@@ -389,7 +389,7 @@ export interface StreamAPI {
   delete(stream: string): Promise<boolean>;
   list(): Lister<StreamInfo>;
   deleteMessage(stream: string, seq: number, erase?: boolean): Promise<boolean>;
-  getMessage(stream: string, seq: number): Promise<StoredMsg>;
+  getMessage(stream: string, query: MsgRequest): Promise<StoredMsg>;
```
ripienaar, line 392: could avoid breaking change with `getLastMessageForSubject()`? if no biggie then lgtm

ripienaar (approved):

nats-base-client/types.ts
aricart: added a shim and deprecation message - the jestream apis in the javascript clients are `preview`

PR #167, opened 2021-06-29
JS add support for `domain`

aricart:
[FEAT] added `domain` to `JetStreamOptions`, when set a prefix `$JS.${domain}.API` is set.

[FEAT] `domain` is reported by `JetStreamAccountStats`


kozlovic (approved):
LGTM

PR #168, opened 2021-06-30
advanced purge operations

aricart:
[FEAT] added the ability to purge streams by subject, sequence, or trimming to a specific number of messages.

### DIFF
--- a/nats-base-client/jsstream_api.ts
+++ b/nats-base-client/jsstream_api.ts
@@ -20,8 +20,11 @@ import {
   MsgDeleteRequest,
   MsgRequest,
   NatsConnection,
+  PurgeBySeq,
+  PurgeBySubject,
   PurgeOpts,
   PurgeResponse,
+  PurgeTrimOpts,
   StoredMsg,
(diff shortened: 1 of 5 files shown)


nats-base-client/jsstream_api.ts
```diff
@@ -85,9 +86,9 @@ export class StreamAPIImpl extends BaseApiClient implements StreamAPI {
     return new ListerImpl<StreamInfo>(subj, filter, this);
   }
 
-  async purge(name: string): Promise<PurgeResponse> {
+  async purge(name: string, opts?: PurgeOpts): Promise<PurgeResponse> {
```
ripienaar, line 89: maybe worth detecting early when there are conflicting values - like you cant have seq and keep together afaik.

aricart: I was hoping the compiler did it, but it looks like they are not there yet.

ripienaar: LGTM

PR #169, opened 2021-06-30
Expose some internal types/headers 

aricart:
[FEAT] Exposed `JsHeaders` enum with possible headers from JetStream (`LastStreamSeqHdr`/`LastConsumerSeqHdr`) which are sent on heartbeats. Also added `StreamSourceHdr`, which is possible when a message is coming from a source.

[FIX] Corrected optional fields from `StreamSource`

[FIX] Exported `isFlowControlMsg`/`isHeartbeatMsg`, `LastForMsgRequest`, `SeqMsgRequest`

[FIX] Fixed a bug when parsing headers that stopped if the header contained a non-default status, non-OK messages may contain additional header entries


kozlovic (approved):
LGTM

PR #178, opened 2021-07-29
[FEAT] support `DeliverPolicy.LastPerSubject`

aricart:
FIX https://github.com/nats-io/nats-architecture-and-design/issues/33

### DIFF
--- a/tests/jetstream_test.ts
+++ b/tests/jetstream_test.ts
@@ -1254,6 +1254,46 @@ Deno.test("jetstream - deliver start time", async () => {
   await cleanup(ns, nc);
 });
 
+Deno.test("jetstream - deliver last per subject", async () => {
+  const { ns, nc } = await setup(jetstreamServerConf({}, true));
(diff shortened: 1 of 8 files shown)


aricart:
Added test conditionals to disable if the feature is not supported.

tests/jetstream_test.ts
```diff
@@ -1254,6 +1254,55 @@ Deno.test("jetstream - deliver start time", async () => {
   await cleanup(ns, nc);
 });
 
+Deno.test("jetstream - deliver last per subject", async () => {
```
kozlovic, line 1257: Not even sure checking against 2.3.3 is enough unless you build from main since the feature is not yet in public release?

kozlovic (approved):
LGTM

tests/jetstream_test.ts
aricart: when I build locally, that works currently reported on local build is 2.3.3-beta or something to that effect.

PR #179, opened 2021-08-04
export required types, update nats-server, kv change

aricart:
- [FIX] exported `PubAck`, `isNatsError()`, `checkJsError()`
- [FIX] JetStream `pull()` tests incorrectly set `expires` as `nanos(expire)` in `PullOptions` - values on the client are expressed as millis, and converted internally. Examples correctly use values as millis
- [UPDATE] nats-server to 2.3.3
- [CHANGE] internal kv implementation changes `keys():Promise<Set<string>>` tp `keys():Promise<string[]>`, this is non-public api.


kozlovic (approved):
LGTM

PR #181, opened 2021-08-16
[feat] expose `PubAck#domain`- where server can distinguish between two same name streams but in different domains.

aricart:
https://github.com/nats-io/nats-architecture-and-design/issues/44

### DIFF
--- a/nats-base-client/types.ts
+++ b/nats-base-client/types.ts
@@ -244,6 +244,7 @@ export interface PullOptions {
 
 export interface PubAck {
   stream: string;
+  domain?: string;
   seq: number;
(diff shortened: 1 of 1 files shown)


kozlovic (approved):
LGTM. Just to make sure: as long as the name of the field matches the name of JSON tag, you have nothing else to do, right?

aricart:
Yes the name matches, I will add a test.

https://github.com/nats-io/nats-server/pull/2432/files#diff-2f4991438bb868a8587303cde9107f83127e88ad70bd19d5c6a31c238a20c299R81

PR #182, opened 2021-08-16
ConsumerConfig model changes

aricart:
model changes SequencePair.last_active, ConsumerInfo#push_bound, ConsumerConfig#deliver_group

https://github.com/nats-io/nats-architecture-and-design/issues/43


kozlovic (approved):
LGTM, I think it. The test is failing because maybe you are running against a server that does not have the deliver group change, so when you expect the failure to be "durable requires queue group 'x'", you get a consumer info from server without DG, so the error returned is "durable requires no queue group".

kozlovic (approved):
LGTM

PR #184, opened 2021-08-17
kv updates - codecs/validation

aricart:
- [feat] added support for key/value codecs, on JavaScript stack the values are always byte arrays, so different codecs are required for keys and values.
- [feat] added validation to keys and buckets

### DIFF
--- a/nats-base-client/kv.ts
+++ b/nats-base-client/kv.ts
@@ -60,6 +60,48 @@ export interface Entry {
   operation: "PUT" | "DEL";
 }
 
+export interface KvCodec<T> {
+  encode(k: T): T;
+  decode(k: T): T;
+}
+
+export interface KvCodecs {
+  key: KvCodec<string>;
+  value: KvCodec<Uint8Array>;
+}
+
+export function Base64KeyCodec(): KvCodec<string> {
+  return {
+    encode(key: string): string {
+      return btoa(key);
+    },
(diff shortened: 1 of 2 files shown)


nats-base-client/kv.ts
```diff
@@ -60,6 +60,48 @@ export interface Entry {
   operation: "PUT" | "DEL";
 }
 
+export interface KvCodec<T> {
```
ripienaar, line 63: in the ADR I mention Encoder, Decoder and Codec seperately because when you are a read only replica client you might not need to supply a Encoder and might even not be able to as you have no private key if all you're supposed to do is reading from a bucket encrypted with a private key. Further reason why I think the read-only variant is important, there's legit cases where even with access to the bucket you couldnt encode data for it.

Your call to support the split, just want to explain why I catered for it.

ripienaar (approved):
LGTM

nats-base-client/kv.ts
aricart: What about if the replica is encrypted?   The providing of the code. Is optional now, but it defaults to noop as a way of insuring all the encode decode paths are covered. I duplicated the crud test with an encoder to verify

ripienaar: Yes, in that case you only need a Decoder not an Encoder.

PR #190, opened 2021-08-27
iterator fixes

aricart:
while writing some tests for KV changes noticed that some unexpected breaks from iterators were leaving behind subscriptions - this PR solves those problems


kozlovic (approved):
LGTM

PR #195, opened 2021-09-08
kv extensions

aricart:
[FEAT] added some kv extensions to delete subkeys as this is super useful to some applications
[FIX] changed kv#keys() to omit keys that have been deleted
[TESTS] Added tests for multiple jetstream pull consumers on the same durable.


kozlovic (approved):
LGTM

PR #196, opened 2021-09-13
Better trace on timeout

aricart:
changed timeout to pre-generate a stack - in the case of a timeout, the stack will have a context to the operation that failed


kozlovic (approved):
LGTM

PR #198, opened 2021-09-14
Kv and cleanup

aricart:
- removed the encoded kv from the lib
- moved types from kv to the types, removed some circ dependencies


kozlovic (approved):
LGTM

PR #201, opened 2021-09-27
FIX partial frames during connect's peek of the info block

aricart:
[fix] `extractProtocolMessage`, was loosely testing the result from `protoLen`, which returned -1. In such a case, it was possible to attempt processing a frame that wasn't complete. This only happened during connect.

### DIFF
--- a/nats-base-client/util.ts
+++ b/nats-base-client/util.ts
@@ -34,13 +34,13 @@ export function protoLen(ba: Uint8Array): number {
       return n + 1;
     }
   }
-  return -1;
+  return 0;
 }
 
 export function extractProtocolMessage(a: Uint8Array): string {
   // protocol messages are ascii, so Uint8Array
   const len = protoLen(a);
-  if (len) {
+  if (len > 0) {
     const ba = new Uint8Array(a);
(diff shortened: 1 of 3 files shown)


nats-base-client/util.ts
```diff
@@ -34,13 +34,13 @@ export function protoLen(ba: Uint8Array): number {
...
   }
-  return -1;
+  return 0;
 }
 
 export function extractProtocolMessage(a: Uint8Array): string {
   // protocol messages are ascii, so Uint8Array
   const len = protoLen(a);
-  if (len) {
+  if (len > 0) {
```
kozlovic, line 43: Since you changed from -1 to 0, the test `if (len) {` would then become valid, but I am ok being explicit here.

kozlovic (approved):
LGTM

nats-base-client/util.ts
aricart: yeap going to keep it that way - at some point I believe the code returned 0, and then changed (this was used by the protocol parser before. Testing for what we are looking for is always a good thing.

PR #202, opened 2021-10-04
[FEAT] dns resolve of server entries

aricart:
Added the ability for client transports to enable dns resolution - currently clients simply pass the server hostname or IP to the network connect function. If specifying `resolve: true` as a connection option, the client will resolve the hostname when attempting to connect to it. If randomize was specified (default), entries are randomized. Note some clients such as nats.ws are unable to resolve, as they WebSocket directly handles the resolution for the connection. If the client fails to resolve any entries, it will throw an error, but the client will attempt to connect to other specified servers if possible.

### DIFF
--- a/nats-base-client/servers.ts
+++ b/nats-base-client/servers.ts
@@ -82,12 +82,14 @@ export class ServerImpl implements Server {
     opts: Partial<{ fn: DnsResolveFn; randomize: boolean; resolve: boolean }>,
   ): Promise<Server[]> {
     if (!opts.fn || !opts.resolve) {
+      // don't add - to resolves or we get a circ reference
       return [this];
     }
 
     const buf: Server[] = [];
     if (isIP(this.hostname)) {
-      buf.push(this);
+      // don't add - to resolves or we get a circ reference
+      return [this];
     } else {
       const ips = await opts.fn(this.hostname);
       for (const ip of ips) {
--- a/nats-base-client/servers.ts
+++ b/nats-base-client/servers.ts
@@ -81,7 +81,8 @@ export class ServerImpl implements Server {
   async resolve(
     opts: Partial<{ fn: DnsResolveFn; randomize: boolean; resolve: boolean }>,
   ): Promise<Server[]> {
-    if (!opts.fn || !opts.resolve) {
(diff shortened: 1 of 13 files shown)


nats-base-client/servers.ts
```diff
@@ -105,6 +117,45 @@ export class Servers {
...
+        }
+        url.hostname = ip;
+        const ss = new ServerImpl(url.host, false);
+        ss.tlsName = s.hostname;
+        buf.push(ss);
+      }
+    }
+    this.servers.length = 0;
+    if (this.randomize) {
+      shuffle(buf);
```
kozlovic, line 154: Is `connect()` invoked only once? If not, are you not possibly re-shuffle the list of URLs that was already shuffled?

Also, remember that the client parity request for host name resolution was that: for a given URL in the list of servers, the library should resolve the hostname and, if getting multiple IPs, shuffle that list, then try to connect with IPs from that list until one succeeds (to TCP connect, not NATS connect).

In other words, suppose the user gives this list of servers as: A, B, C. Say library shuffled that to: C, A, B.
When connecting, you do: use "C", it is a host name, resolve, say you get IP_C only. Use that one. If fail to connect, move to A, say resolution returns IP_A1, IP_A2, IP_A3, shuffle that list, then use the first, if it TCP connects OK, then continue connection process, otherwise, go to the next in the "IP_A" list, etc..
But for the "servers" list, the "C, A, B" is shuffled once, then it is a matter of moving the tried server to the back. After trying "C", the list becomes "A, B, C", then after trying "A", "B, C, A", etc..

aricart: The updated implementation does exactly this. It is actually simpler.

kozlovic:
As discussed, I would make it default, and not an option to disable (for now). There is an issue of stack overflow that is unexplained so far, and could be a deno API bug so question is if this feature needs to be disabled for now on deno (it is already on ws).

aricart:
@kozlovic tracked the issue as I mentioned and resolved it - also removed the option, so by default if the transport supports resolve, it will just happen.

kozlovic (approved):
LGTM

PR #208, opened 2021-10-13
[FIX] flow control/protocol messages now handled when they would be dispatched.

aricart:
The queued iterator has a filterFn that tests the inbound message to see if it should be presented to the callback/iterator.
This enables the processing of the flow control messages when the user would have seen it, rather than when the protocol injected it into the subscription.


kozlovic (approved):
LGTM

PR #211, opened 2021-10-18
Exposed newly added consumer/stream options

aricart:
[FEAT] enhanced ConsumerOptsBuilder to expose same options as Go implementation

[FEAT] exposed support for stream options `sealed`, `deny_delete`, `deny_purge`, `allow_rollup_hdrs`

[FEAT] rollup can be configured by setting the header `JsHeaders.RollupHdr` with a value `JsHeaders.RollupValueSubject` or `JsHeaders.RollupValueAll`.

[FEAT] exposed `ConsumerConfig.headers_only` - enabling a consumer to retrieve messages without including the payload. To see the size of the payload, inspect the `JsHeaders.MessageSizeHdr`.

### DIFF
--- a/tests/jsm_test.ts
+++ b/tests/jsm_test.ts
@@ -216,7 +216,6 @@ Deno.test("jsm - add stream", async () => {
   assertEquals(si.config.name, name);
 
   const fn = (i: StreamInfo): boolean => {
-    console.log(i.config, si.config);
     assertEquals(i.config, si.config);
     assertEquals(i.state, si.state);
     assertEquals(i.created, si.created);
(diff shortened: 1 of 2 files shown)


tests/jsm_test.ts
```diff
@@ -216,6 +216,7 @@ Deno.test("jsm - add stream", async () => {
   assertEquals(si.config.name, name);
 
   const fn = (i: StreamInfo): boolean => {
+    console.log(i.config, si.config);
```
matthiashanel, line 219: the log statement intentional?

PR #216, opened 2021-11-02
request validation doesn't reject properly in downstream

aricart:
[FIX] in some downstream environments the assertion for the subject doesn't seem to auto-reject properly even if there's a handler associated with the promise.


kozlovic (approved):
LGTM

PR #227, opened 2021-11-11
[FEAT] surface JS errors as per ADR-7

### LINKED ISSUE #226: Provide detailed error details in exceptions
I'm using version `2.2.0` of the Javascript client and a call to `    jsm = await nc.jetstreamManager()` throws an exception.  The details contained in the Error object that is thown is `{"name":"NatsError","code":"503"}`.  Looking at the error message provided by the nats-server, there is a lot more information that should be bubbled up to the application via the Error object.   Here is the info provided by the server: 
```
{"code":503,"err_code":10008,"description":"JetStream system temporarily unavailable"}
```

The detailed error code `10008` as well as the description should be available.  Error code `503` alone is not enough to troubleshoot.

aricart:
The modification enhances NatsError by adding:

`isJetStreamError(): boolean`
`jsError(): ApiError|null`

This enables clients to test if there's more info to the underlying error

FIX #226

### DIFF
--- a/nats-base-client/error.ts
+++ b/nats-base-client/error.ts
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
+import { ApiError } from "./types.ts";
+
 export enum ErrorCode {
   // emitted by the client
   ApiError = "BAD API",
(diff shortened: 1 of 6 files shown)


kozlovic (approved):
LGTM

PR #229, opened 2021-11-29
JSM consumer `update`

aricart:
added a typed consumer `update` - while JetStream simply does an `add`, the update api, simply retrieves the consumer, and assigns values present in the update (typed to show what properties are editable) - note that this is only intended for updates to durable consumers.

Modifying a consumer is only supported on nats-server 2.6.4+


ripienaar (approved):
LGTM

PR #230, opened 2021-11-29
typed stream `update`

aricart:
this creates a breaking change - for API symmetry with update consumer, the name of the stream is now a required argument.


aricart:
@kozlovic the change for the consumer is here:

https://github.com/nats-io/nats.deno/pull/229

kozlovic (approved):
LGTM

PR #232, opened 2021-11-30
Deno 1.16.x

aricart:
[CHANGE] [DENO] internals for handling signals in the test framework upgraded to changes from Deno.Signal
[CHANGE] [DENO] tls upgrade changed name of CA property - and instead of a file, takes a string
[BREAKING] [DENO] after this change only Deno 1.16.x or better is compatible with the library.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
     environment: CI
     strategy:
       matrix:
-        deno-version: [1.8.3, 1.9.2, 1.13.2]
+        deno-version: [1.16.3]
 
     steps:
       - name: Git Checkout Deno Module
--- a/examples/bench.js
+++ b/examples/bench.js
@@ -81,8 +81,8 @@ const reducer = (a, m) => {
     a.version = m.version;
     a.async = m.async;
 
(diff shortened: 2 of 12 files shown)


kozlovic (approved):
LGTM

dontlaugh:
I think I get a related error when running my build against nats.deno version 1.4.0

```
error: TS2345 [ERROR]: Argument of type '{ hostname: string; certFile: string | undefined; }' is not assignable to parameter of type 'StartTlsOptions'.
  Object literal may only specify known properties, and 'certFile' does not exist in type 'StartTlsOptions'.
      { hostname, certFile: tls.caFile },
```

Deno version

```
% deno --version
deno 1.16.3 (release, x86_64-unknown-linux-gnu)
v8 9.7.106.5
typescript 4.4.2
```

My build script for our program (bs.ts)

```bash
cmd_build () {
    DENO_DIR=./cache deno cache --unstable --lock=lock.json --lock-write bs.ts
}
```

Using the raw github import fixes my issue.

```
import * as nats from "https://raw.githubusercontent.com/nats-io/nats.deno/main/src/mod.ts";

```

PR #233, opened 2021-12-01
small fixes to KV

aricart:
[FEAT] KvEntry now adds `length` to telegraph the length of the payload - if `headers_only` this value is extracted from the header sent by the server, otherwise, it is calculated from the payload.

[FIX] KvOption didn't expose the storage type for the KV - it is now possible to specify `StorageType.File` (the default) or `StorageType.Memory`.

[FEAT] history() and watch() can now specify `headers_only` as an option.


kozlovic (approved):
LGTM

PR #238, opened 2021-12-20
[FIX] replaced fast encode/decode with TextEncoder/Decoder

### LINKED ISSUE #237: Username/password authenticator password encoding issue
## Reproduce

1. Run a nats server with the following configuration:
```
listen: 0.0.0.0:4222

authorization {
    username: admin
    password: §12§12§12
}

websocket {
    listen: 0.0.0.0:443
    no_tls: true
}
```
2. Connect to the server with the ws client usgin username/password auth

### Expected result

The client can authenticate successfully

### Effective result

The client receives an "Authentication Violation" error.

## Debug

When running a custom server and registering a custom authenticator, we can print the received password and we notice that the encoding of `§` character is wrong.

## Workaround

Encode the password in base64 when passing it to the connect function. Create a custom authenticator on the server side and decode the password.

aricart:
[FIX] fixed an issue where the encoding data was not working correctly as the base library was using a fast encoding/decoding algorithm which doesn't work correctly for higher UTF code-points. The fix simply delegates to standard TextEncoder/TextDecoder.

FIX #237

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
     environment: CI
     strategy:
       matrix:
-        deno-version: [1.16.3]
+        deno-version: [1.17.0]
 
     steps:
       - name: Git Checkout Deno Module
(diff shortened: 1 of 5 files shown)


kozlovic (approved):
LGTM

PR #239, opened 2021-12-23
KV prefix handling

aricart:
[FEAT] Properly support KV across accounts:

a sample server configuration:

```javascript
accounts: {
      A: {
        jetstream: true,
        users: [{ user: "a", password: "a" }],
        exports: [
          { service: "$JS.API.>" },
          { service: "$KV.>" },
          { stream: "forb.>" },
        ],
      },
      B: {
        users: [{ user: "b", password: "b" }],
        imports: [
          { service: { subject: "$JS.API.>", account: "A" }, to: "froma.>" },
          { service: { subject: "$KV.>", account: "A" }, to: "froma.$KV.>" },
          { stream: { subject: "forb.>", account: "A" } },
        ],
      },
    },
```
In the above:

- A exports:
  -  Its JetStream API  - required to use KV
  - A service all subjects relating to KV
  - A stream
- B imports:
  - The JetStream API
  - The KV subjects, using a prefix
  - The prefixed stream

In ordered to access the KV, the client in `B`:
- Creates a connection setting the `inboxPrefix` to `forb` (this will force NATS to know that subscription interest on `forb.>` is required by account `B` and allow the ordered push consumer to work for watches and history.
- Using the same connection, the client creates a `JetStreamClient`, setting the option `apiPrefix` to `froma`, `JetStreamManager` requests to setup the KV, etc, will now flow to the account's `A` JetStream.
- KVs created with that JetStream will now properly target buckets in `A`.

[CHANGE] internal Bucket.create now requires a JetStreamClient instead of a NatsConnection - this allows propagation of the JetStream options to the KV, which enables prefixes (from the NatsConnection) and `apiPrefix` from the `JetStreamClient` to be properly honored by the Kv.

[FIX] consumer options ignored `inboxPrefix` connection option when auto-generating an inbox for a `deliverTo`, this, in turn, prevented things like watch/history from working on across accounts because in order to circumvent consumer issues across accounts a stream must be exported, and subscriber must use this prefix in order to receive messages. Possibly this is incorrect but requires current consumer across account strategy to be changed to pull (or at least support ordered pull consumers) 

[FIX] when re-creating an ordered consumer, the API prefix was incorrectly used as an inbox prefix.

### DIFF
--- a/nats-base-client/jsclient.ts
+++ b/nats-base-client/jsclient.ts
@@ -398,7 +398,7 @@ export class JetStreamClientImpl extends BaseApiClient
           ErrorCode.ApiError,
         );
       }
-      jsi.config.deliver_subject = createInbox();
+      jsi.config.deliver_subject = createInbox(this.nc.options.inboxPrefix);
       jsi.config.ack_policy = AckPolicy.None;
       jsi.config.max_deliver = 1;
       jsi.config.flow_control = true;
@@ -560,7 +560,7 @@ class JetStreamSubscriptionImpl extends TypedSubscription<JsMsg>
     if (this.info === null || this.sub.isClosed()) {
       return;
     }
-    const newDeliver = createInbox(this.js.opts.apiPrefix);
+    const newDeliver = createInbox(this.js.nc.options.inboxPrefix);
     const nci = this.js.nc;
     nci._resub(this.sub, newDeliver);
     const info = this.info;
--- a/nats-base-client/jsclient.ts
+++ b/nats-base-client/jsclient.ts
@@ -89,8 +89,8 @@ class ViewsImpl implements Views {
     this.js = js;
     jetstreamPreview(this.js.nc);
   }
-  async kv(name: string, opts: Partial<KvOptions> = {}): Promise<KV> {
-    return Bucket.create(this.js.nc, name, opts);
+  kv(name: string, opts: Partial<KvOptions> = {}): Promise<KV> {
+    return Bucket.create(this.js, name, opts);
   }
 }
 
--- a/nats-base-client/kv.ts
+++ b/nats-base-client/kv.ts
@@ -228,7 +228,7 @@ export class Bucket implements KV, KvRemove {
     return `${this.subjPrefix}.${this.bucket}.${k}`;
   }
 
-  getSubjectForKey(k: string): string {
+  fullKeyName(k: string): string {
     return `${kvSubjectPrefix}.${this.bucket}.${k}`;
   }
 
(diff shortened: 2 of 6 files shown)


nats-base-client/jsclient.ts
```diff
@@ -398,7 +398,7 @@ export class JetStreamClientImpl extends BaseApiClient
           ErrorCode.ApiError,
         );
       }
-      jsi.config.deliver_subject = createInbox();
+      jsi.config.deliver_subject = createInbox(this.nc.options.inboxPrefix);
```
matthiashanel, line 401: so `this.nc.` does not have a create inbox function that makes use of the inbox Prefix automatically?
As is the case in go.

nats-base-client/kv.ts
```diff
@@ -158,24 +156,35 @@ export class Bucket implements KV, KvRemove {
...
+    const jsi = js as JetStreamClientImpl;
+    const prefix = jsi.prefix || "$JS.API";
+    if (prefix !== "$JS.API") {
+      this.subjPrefix = `${prefix}.${kvSubjectPrefix}`;
+    }
   }
 
   static async create(
-    nc: NatsConnection,
+    js: JetStreamClient,
```
matthiashanel, line 177: I am a bit confused between passing in the JetStreamClient and creating the jetstream manager locally. 
Would passing in the options or the manager by any better?

nats-base-client/kv.ts
```diff
@@ -212,10 +221,14 @@ export class Bucket implements KV, KvRemove {
...
   subjectForBucket(): string {
-    return `${kvSubjectPrefix}.${this.bucket}.>`;
+    return `${this.subjPrefix}.${this.bucket}.>`;
   }
 
   subjectForKey(k: string): string {
+    return `${this.subjPrefix}.${this.bucket}.${k}`;
+  }
+
+  getSubjectForKey(k: string): string {
```
matthiashanel, line 231: maybe pick a different name that calls out the difference.
`get` is a bit non descript.

matthiashanel (commented):
Just a few questions

nats-base-client/jsclient.ts
aricart: no it doesn't - it is just a function. But anywhere it is called in the lib it is honored.

nats-base-client/kv.ts
aricart: that is just internal - the clients get it as nc.jetstream().views.kv(name)

matthiashanel: ok

matthiashanel (approved):
LGTM

PR #248, opened 2022-01-22
[FIX] processOptions processed `port` property before `servers`

### LINKED ISSUE #247: port and servers connect options conflict
The docs says the port is only used if no servers provided but in the options.js the opts.port is checked first and servers is set with default host and port.
This is version 2.5.0

function parseOptions(opts) {
    const dhp = `${types_1.DEFAULT_HOST}:${(0, transport_1.defaultPort)()}`;
    opts = opts || { servers: [dhp] };
    if (opts.port) {
        opts.servers = [`${types_1.DEFAULT_HOST}:${opts.port}`];
    }
    if (typeof opts.servers === "string") {

aricart:
If an ambiguous configuration was provided (one that specified `port` and `servers`), the `port` option would be selected, discarding the specified `servers`. The documentation for the JavaScript clients specifies that port is only processed if no servers are specified.

FIX #247


kozlovic (approved):
LGTM

PR #249, opened 2022-01-25
Fixes associated with NATS server 2.7 behavior changes

aricart:
[FIX] nats-server 2.7+ handle mixes of `no_wait` and `expires` are slightly different. Since `no_wait` is exposed, the javascript client handles that option by eliminating any specification of `expires`.

[FIX] nats-server 2.7+ returns a 408 when `expires` triggers - previously it was a NoMessages, the 408 is remapped to the no messages which then is handled properly.

[CHANGE] nats-server 2.7+ allows pull consumers to be ephemeral, added a fix for the determination of the consumer `name`, since previously it was the durable, but now it is the consumer name returned by the create consumer.

### DIFF
--- a/nats-base-client/jsutil.ts
+++ b/nats-base-client/jsutil.ts
@@ -96,7 +96,7 @@ export function checkJsErrorCode(
   switch (code) {
     case 408:
       return NatsError.errorForCode(
-        ErrorCode.JetStream404NoMessages,
+        ErrorCode.JetStream408RequestTimeout,
         new Error(description),
       );
     case 503:
--- a/nats-base-client/jsutil.ts
+++ b/nats-base-client/jsutil.ts
@@ -94,6 +94,11 @@ export function checkJsErrorCode(
   }
   description = description.toLowerCase();
   switch (code) {
+    case 408:
+      return NatsError.errorForCode(
+        ErrorCode.JetStream404NoMessages,
+        new Error(description),
+      );
     case 503:
       return NatsError.errorForCode(
         ErrorCode.JetStreamNotEnabled,
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,7 +30,7 @@ jobs:
           deno-version: ${{ matrix.deno-version }}
 
       - name: Set NATS Server Version
-        run: echo "NATS_VERSION=v2.6.5" >> $GITHUB_ENV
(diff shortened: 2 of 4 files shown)


nats-base-client/jsutil.ts
```diff
@@ -94,6 +94,11 @@ export function checkJsErrorCode(
   }
   description = description.toLowerCase();
   switch (code) {
+    case 408:
+      return NatsError.errorForCode(
+        ErrorCode.JetStream404NoMessages,
```
kozlovic, line 99: A 408 is not a "no message", it is a request timeout or no longer valid, but not a "no message"

kozlovic (changes requested):

nats-base-client/jsutil.ts
aricart: the server is sending me that in response...

scottf:
@kozlovic Are there plans to update ADR's to cover this?

kozlovic:
@scottf cover what? There have always been several possible codes returned by the server, this is not new.

scottf:
> @scottf cover what? There have always been several possible codes returned by the server, this is not new.

The changes to pull behavior and api/data. Changes in v2.7.0 broke the Java client unit tests, which I only noticed today

kozlovic:
> The changes to pull behavior and api/data. Changes in v2.7.0 broke the Java client unit tests, which I only noticed today

When v2.7.0 was released, we did mention that the server would now allow pull consumers with ephemeral and have added options (don't have to use them, but they are there). From release notes, in the Added section:
```
- Support for ephemeral Pull consumers (client libraries will need to be updated to allow those) (#2776)
- New consumer configuration options (#2776):
    - For Pull Consumers:
        - MaxRequestBatch to limit the batch size any client can request
        - MaxRequestExpires to limit the expiration any client can request
    - For ephemeral consumers :
        - InactiveThreshold duration that instructs the server to cleanup ephemeral consumers that are inactive for that long
```
In the Improved section:
```
Pull consumers behavior: more resilient, better at waiting request management, and the ability to determine and watch interest cross account (#2776)
```
I did not make those changes, simply wrote the release notes. I know that some tests in the Go/C client failed because for instance I was checking for pull consumer requiring a durable name, which is no longer required.

A part from that, the pull consumer API should not "break" with v2.7.0+.

kozlovic (approved):
LGTM

PR #254, opened 2022-01-30
hostPort fixes for http/s protocol default parsing

aricart:
parsing the hostPort will default the port to 4222 if the platform's URL#port doesn't report default ports for a protocol even if specified in the URL (like 80 or 443).

FIX https://github.com/nats-io/nats.js/issues/478

### DIFF
--- a/nats-base-client/servers.ts
+++ b/nats-base-client/servers.ts
@@ -25,13 +25,14 @@ import { defaultPort, getUrlParseFn } from "./transport.ts";
 import { shuffle } from "./util.ts";
 import { isIP } from "./ipparser.ts";
 
-function isIPV4OrHostname(hp: string): boolean {
+export function isIPV4OrHostname(hp: string): boolean {
   if (hp.indexOf(".") !== -1) {
     return true;
   }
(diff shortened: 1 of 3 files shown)


nats-base-client/servers.ts
```diff
@@ -36,12 +37,20 @@ function hostPort(
   // that means that protocols other than HTTP/S are not
   // parsable correctly.
   const url = new URL(`http://${u}`);
+
+  let sport = "";
   if (!url.port) {
-    url.port = `${DEFAULT_PORT}`;
+    if (u.endsWith(":80")) {
```
kozlovic, line 43: Is there not a `URL` function that parses host and port for you? It looks strange here that you look for :80 or :443. I would simply try to parse the port, and if there is no port, use the DEFAULT_PORT.
Note that what we have done for websocket in the Go library is to say that if protocol is "ws" then default port is 80, and 443 for wws, any other is 4222.

aricart: Yes no such thing - the websocket default ports etc are handled in the websocket library. There will be some work there as well.

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #255, opened 2022-02-01
[FIX] added constraint to reject options when both servers and port are specified.

aricart:
[FIX] added check that either servers or port was provided - if both options are provided the client will throw an error, as that creates ambiguity on the purpose of `port` - `port` is only allowed if no servers are specified.

FIX https://github.com/nats-io/nats.js/issues/479


kozlovic (approved):
LGTM

Issue #256, opened 2022-02-09
JetStreamManager.update() requires stream name in cfg object?


JH7:
Calling the JetStreamManager.update() method like this (using the nats.js in node.js):
```js
jcm.streams.update('test', options);
```
throws the error:
```js
{"name":"NatsError","code":"500","api_error":{"code":500,"err_code":10052,"description":"stream name is required and can not contain '.', '*', '>'"}}
```

My guess is that the stream name has to be included in the `cfg` object used here:
https://github.com/nats-io/nats.deno/blob/d698a5ad688bba907fc2cdb962b360bc75ff6fd3/nats-base-client/jsmstream_api.ts#L78-L81

If I manually include the name like so `jcm.streams.update('test', { name: 'test', ...options});`, the error is not thrown.

Additionally, the TypeScript declaration of `StreamUpdateConfig` does not include `subjects`, even though I was able to call the `update()` method with a `cfg` object including `subjects` and the stream seemed to be updated correctly.
https://github.com/nats-io/nats.deno/blob/5bf5482cd2e62a12e8a4dfe350624794491a0254/nats-base-client/types.ts#L560-L572
Furthermore: Are `max_msgs`, `max_age` and `max_bytes` really mandatory for an update (as shown in the TS interface)?

I'm not into the server/protocol implementation of NATS, so I currently cannot check whether my assumptions are actually true or not, so apologies if I'm missing something.

aricart:
The incorrect thing is: `update(name: string, cfg: StreamUpdateConfig): Promise<StreamInfo>` - this should be:
`update(name: string, cfg: Partial<StreamUpdateConfig>): Promise<StreamInfo>` which will relax all fields except for the ones you are changing. The library can inject the `name` for you as that is provided as a separate argument.

JH7:
> @JH7 what server/client version are you on?

Server: 2.7.2
Client (nats): 2.6.0

aricart:
fixed by https://github.com/nats-io/nats.deno/pull/258

PR #258, opened 2022-02-10
Fixes to `StreamUpdateConfig`, `StreamConfig` and jsm.streams.update() signature

aricart:
[FIX] StreamUpdateConfig added `subjects`, also unmarked optional `max_msgs_per_subject`, `max_msg_size`, `discard`, and `duplicate_window` - The update api, makes all these fields optional as it is a partial, but when the StreamConfig is returned, these values will be set to their server default values (thus expected to be set).

[FIX] StreamConfig removes `subjects` as it is provided by the extended interface of StreamUpdateConfig

[CHANGE] jsm.streams.update now takes a Partial<StreamUpdateConfig> - this relaxes all fields to be optional

[FIX] jsm.stream.update(name, Partial<StreamUpdateConfig>) will now set the name field - if the update was done without an initial copy of the stream config, this value might not be set by default and result in the server returning an error

### DIFF
--- a/nats-base-client/jsmstream_api.ts
+++ b/nats-base-client/jsmstream_api.ts
@@ -64,7 +64,7 @@ export class StreamAPIImpl extends BaseApiClient implements StreamAPI {
 
   async update(
     name: string,
-    cfg = {} as StreamUpdateConfig,
+    cfg = {} as Partial<StreamUpdateConfig>,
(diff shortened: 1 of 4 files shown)


aricart:
FIX #256

kozlovic (approved):
LGTM

PR #259, opened 2022-02-14
[FIX] node v17.5.0 seems to go in a loop, suspect some internal patch of subarray to slice

aricart:
In node, the nbc library conditionally patched buf.subarray as buf.slice because in node, subarray performed copies of the data. This change removes this optimization (as node 17.5.0 seems to internally alias it and thus create an allocation loop which exhausts the stack space).

FIX https://github.com/nats-io/nats.js/issues/483


kozlovic (approved):
LGTM

PR #260, opened 2022-02-16
server version

aricart:
added internal support to get the semver of the connection server - this will be required for some JetStream operations.


kozlovic (approved):
LGTM

PR #262, opened 2022-02-17
KV watch notifications and discard policy

aricart:
[FEAT] kv watch now takes a callback that will get called once the entries existing in the kv have been received, this allows for the client to get a notification of when the existing values have been loaded (watch will continue to run, notifying of other changes)

[CHANGE] kv will now use DiscardPolicy.New when creating kvs on servers 2.7.2 or better

Fixes https://github.com/nats-io/nats-architecture-and-design/issues/94


kozlovic (approved):
LGTM

PR #267, opened 2022-02-18
FIX 265

### LINKED ISSUE #265: javascript clients close the connection on permission error
The JavaScript client since 2.0 deviates from standard practice of allowing clients to keep connected when pub/sub pub/sub permissions are raised.

aricart:
[FIX] [CHANGE] current javascript clients close the connection on permission errors, this changes the behaviour to match other client implementations. Note that in some cases (like pub/req permissions) the only way to know that there was a permission error is via the status. On subscriptions, if callback, the `err` argument will provide the same.

[FIX] on iterators, if a permission error triggered, the connection was closed, but the promise from the subscription backing it, remained pending.

Fixes #265


Jarema (approved):
LGTM

PR #282, opened 2022-04-14
JSMsg.next() - pull next message (or batch)

aricart:
[FEAT] next() - allows a JetStream pull subscription using a callback to simplify ack'ing and pulling the next message (or batch of messages). This strategy doesn't require timers, as the server returns an error if the request times out.


kozlovic (approved):
LGTM!

Issue #279, opened 2022-04-11
Support chaining on ConsumerOptionsBuilder


avin-kavish:
Would be good to have chaining support on the builder:

```ts
const opts = consumerOpts().durable(...).ackExplicit()
```

Would you accept a PR for this?

aricart:
@avin-kavish sure!

PR #283, opened 2022-04-14
Allow chaining consumer options builder

avin-kavish:


aricart:
Thank you @avin-kavish!

aricart:
FIXES #279

PR #287, opened 2022-05-02
export token/usernamePassword authenticators

### LINKED ISSUE #275: token authenticator function is not exposed on the client api
https://github.com/nats-io/nats.deno/blob/main/nats-base-client/authenticator.ts#L22

aricart:
[FIX] exported the various `Authenticator` return types
[FIX] exported functions `tokenAuthenticator` and `usernamePasswordAuthenticator`

FIX #275


kozlovic (approved):
LGTM

PR #288, opened 2022-05-03
Fixes related to JetStream errors being prematurely processed by the core request layer

### LINKED ISSUE #273: NatsError `code` is concat of code + message?
# Context

I have a pull consumer, to which I am calling single `js.pull()` s.  From which I am attempting to catch the 404, 408 and 409 Jetstream errors to then "sleep" when no messages come back.

# Issue

Examining `err.code` does not match any values in the `ErrorCode` enum.
Instead, the `code` field looks to be a concatenation of the enum code and a text message.

Example:

I'm expecting `404` as the `nats.ErrorCode.JetStream404NoMessages` code value, but i'm getting `404 No Messages` ...

# Sample Code

```
  import { nats } from "./deps.ts";

  ... [omitted class code]...

  async pull(): Promise<nats.JsMsg> {
    while (true) {
      try {
        console.log("awaiting");
        const msg = await this.js.pull(this.streamName, this.consumerName);
        return msg;
      } catch (err) {
        if (
          err.code == nats.ErrorCode.Timeout ||
          err.code == nats.ErrorCode.JetStream404NoMessages ||
          err.code == nats.ErrorCode.JetStream408RequestTimeout ||
          err.code == nats.ErrorCode.JetStream409MaxAckPendingExceeded
        ) {
          // Sleep Pause
          console.log(
            `JSPULL: ${nats.ErrorCode}; Sleep ${this.pollWaitInterval} ms`,
          );
          await new Promise((resolve) =>
            setTimeout(resolve, this.pollWaitInterval)
          );
        } else {
          console.log(`throwing??? ${err.code}`);
          throw err;
        }
      }
    }
  }
```

# Environment

`deps.ts`

```
export * as nats from "https://deno.land/x/nats@v1.6.1/src/mod.ts";
export * as oak from "https://deno.land/x/oak@v10.4.0/mod.ts";
export * as prom from "https://deno.land/x/ts_prometheus@v0.3.0/mod.ts";
```

`Dockerfile`

```
FROM denoland/deno:1.19.3

WORKDIR /app

USER deno

ADD . .
RUN deno cache --unstable main.ts

EXPOSE 9091

CMD ["run", "--allow-net", "--allow-env", "--unstable", "main.ts"]
```

Initialization:

```

// We init the connection options, and force infinite reconnect attempts
const copts = { servers: NATS_SERVERS } as nats.ConnectionOptions;
copts.inboxPrefix = NATS_INBOX_PREFIX;
copts.debug = NATS_DEBUG == "true" ? true : false;
copts.maxReconnectAttempts = -1;
copts.reconnect = true;

// We use the nkey+jwt credentials data if available
if (NATS_CREDENTIALS) {
  copts.authenticator = nats.credsAuthenticator(
    new TextEncoder().encode(NATS_CREDENTIALS),
  );
}

// Connect to the NATS server. On failure to first connect,
// this will throw and error and stop the deno vm.
const nc = await nats.connect(copts);
const jsopts = {} as nats.JetStreamOptions;
jsopts.timeout = 10000;
const js = nc.jetstream(jsopts);
```

aricart:
[FIX] the base client request introspected a message and if an status code was set, it turned that into an error. The logic for this was not quite correct one requirement is for the message payload of this type of error be empty and NATS core currently can only expect error of 503 (no responders). Any interpretation of the status code unless a 503, should be delegated - in this case to JetStream which can then make better interpretation of the code (which could simply be a marker for no messages, or a request timeout on requests that have an expiration.

[CHANGE] js.pull(stream,dur) now also adds an optional expires - the expires also overrides the request timeout.

FIX #273

### DIFF
--- a/nats-base-client/jsutil.ts
+++ b/nats-base-client/jsutil.ts
@@ -66,6 +66,9 @@ export function millis(ns: Nanos) {
 }
 
 export function isFlowControlMsg(msg: Msg): boolean {
+  if (msg.data.length > 0) {
+    return false;
+  }
   const h = msg.headers;
   if (!h) {
     return false;
@@ -78,6 +81,10 @@ export function isHeartbeatMsg(msg: Msg): boolean {
 }
 
 export function checkJsError(msg: Msg): NatsError | null {
+  // JS error only if no payload - otherwise assume it is application data
+  if (msg.data.length !== 0) {
+    return null;
+  }
   const h = msg.headers;
   if (!h) {
     return null;
@@ -94,9 +101,15 @@ export function checkJsErrorCode(
   }
   description = description.toLowerCase();
   switch (code) {
+    case 404:
+      // 404 for jetstream will provide different messages ensure we
+      // keep whatever the server returned
+      return new NatsError(description, "404");
     case 408:
+      return new NatsError(description, ErrorCode.JetStream408RequestTimeout);
+    case 409:
       return NatsError.errorForCode(
-        ErrorCode.JetStream408RequestTimeout,
+        ErrorCode.JetStream409MaxAckPendingExceeded,
         new Error(description),
       );
     case 503:
--- a/nats-base-client/jsutil.ts
+++ b/nats-base-client/jsutil.ts
@@ -111,7 +111,7 @@ export function checkJsErrorCode(
       // the description can be exceeded max waiting or max ack pending
       return new NatsError(
         description,
-        ErrorCode.JetStream409MaxWaitingExceeded,
+        ErrorCode.JetStream409,
       );
     case 503:
       return NatsError.errorForCode(
--- a/nats-base-client/jsutil.ts
+++ b/nats-base-client/jsutil.ts
@@ -104,13 +104,14 @@ export function checkJsErrorCode(
     case 404:
       // 404 for jetstream will provide different messages ensure we
       // keep whatever the server returned
-      return new NatsError(description, "404");
+      return new NatsError(description, ErrorCode.JetStream404NoMessages);
     case 408:
       return new NatsError(description, ErrorCode.JetStream408RequestTimeout);
     case 409:
-      return NatsError.errorForCode(
-        ErrorCode.JetStream409MaxAckPendingExceeded,
-        new Error(description),
+      // the description can be exceeded max waiting or max ack pending
+      return new NatsError(
+        description,
+        ErrorCode.JetStream409MaxWaitingExceeded,
       );
     case 503:
       return NatsError.errorForCode(
--- a/tests/jetstream_test.ts
+++ b/tests/jetstream_test.ts
@@ -3080,7 +3080,6 @@ Deno.test("jetstream - pull error: max_waiting", async () => {
       3000,
       ErrorCode.JetStream408RequestTimeout,
     ),
-    // this a server bug, it should be a 408
     expectError(3000, ErrorCode.JetStream409MaxWaitingExceeded),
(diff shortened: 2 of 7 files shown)


tests/jetstream_test.ts
```diff
@@ -3010,3 +3010,99 @@ Deno.test("jetstream - pull next", async () => {
...
+      if (message) {
+        assertEquals(err.message, message);
+      }
+    }
+  }
+  await Promise.all([
+    expectError(
+      3000,
+      ErrorCode.JetStream408RequestTimeout,
+      "408 request canceled",
```
kozlovic, line 3083: So as of server rev (https://github.com/nats-io/nats-server/commit/cadf921ed162951e69b6a031648b107c83f63241), when you reach the maximum number of pull requests, the new request will get a 409 (Exceeded MaxWaiting), and not an old one getting a 408 (Request Canceled). This was incorrect behavior. The server release notes for the upcoming 2.8.2 will state:
```
### Changed
- JetStream:
    - When reaching the max number of pull requests, the server would
      evict existing pending requests (with a `408 Request Canceled`) 
      instead of sending a `409 Exceeded MaxWaiting` to the new request (#3099)
```

kozlovic (changes requested):
You may want to change your test if you are going to run against the server from the main branch...

nats-base-client/jsutil.ts
```diff
@@ -94,10 +101,17 @@ export function checkJsErrorCode(
...
     case 408:
-      return NatsError.errorForCode(
-        ErrorCode.JetStream408RequestTimeout,
-        new Error(description),
+      return new NatsError(description, ErrorCode.JetStream408RequestTimeout);
+    case 409:
+      // the description can be exceeded max waiting or max ack pending
+      return new NatsError(
+        description,
+        ErrorCode.JetStream409MaxWaitingExceeded,
```
kozlovic, line 114: Why name you "409" status JetStream409MaxWaitingExceeded or JetStream409MaxAckPendingExceeded? Should it not be simply JetStream409 or JetStream409Status? The description is then used to get a better sense of what the reason was...

tests/jetstream_test.ts
```diff
@@ -3010,3 +3010,97 @@ Deno.test("jetstream - pull next", async () => {
...
+      assertEquals(err.code, code);
+    }
+    return d;
+  }
+  const errors = await Promise.all([
+    expectError(
+      3000,
+      ErrorCode.JetStream408RequestTimeout,
+    ),
+    // this a server bug, it should be a 408
```
kozlovic, line 3083: Why? You have the max number of pull requests waiting, so why is it a server bug to send a 409 to indicate that the max number of pulls is reached? Or are you saying that it should be a 408 Exceeded MaxWaiting instead of 409 Exceeded MaxWaiting?

aricart: The server was canceling the first request (the one that was valid) and then hanging on the second. I removed the comment - I am OK with the current behaviour.

kozlovic: The hanging was the bug with MaxWaiting==1 (which is now fixed). While fixing that issue, we also fixed the fact that a new request on max_waiting reached should fail instead of evicting a current one.

nats-base-client/jsutil.ts
aricart: API breakage --- those are exported, and we added new things. But perhaps we can deprecate the old. Let's do that.

kozlovic (approved):
LGTM

PR #289, opened 2022-05-04
FIX unhandled promises on transport.send()

### LINKED ISSUE #271: The network exception is not handled normally
An exception may be thrown here, but it is not handled correctly:
https://github.com/nats-io/nats.js/blob/main/src/node_transport.ts#L332
https://github.com/nats-io/nats.deno/blob/main/nats-base-client/transport.ts#L76
https://github.com/nats-io/nats.deno/blob/main/nats-base-client/protocol.ts#L725

```
send error:  Error: write EPIPE
    at afterWriteDispatched (internal/stream_base_commons.js:154:25)
    at writeGeneric (internal/stream_base_commons.js:145:3)
    at Socket._writeGeneric (net.js:782:11)
    at Socket._write (net.js:794:8)
    at writeOrBuffer (_stream_writable.js:352:12)
    at Socket.Writable.write (_stream_writable.js:303:10)
    at NodeTransport.send (/Users/Shonke/Developer/nats-1/node_modules/nats/src/node_transport.js:275:21)
    at ProtocolHandler.flushPending (/Users/Shonke/Developer/nats-1/node_modules/nats/nats-base-client/protocol.js:554:28)
    at Timeout._onTimeout (/Users/Shonke/Developer/nats-1/node_modules/nats/nats-base-client/protocol.js:388:22)
    at listOnTimeout (internal/timers.js:549:17) {
  errno: -32,
  code: 'EPIPE',
  syscall: 'write'
}
```

aricart:
These could bubble up and be trapped by the runtime, resulting in the client crashing.

FIXES #271

### DIFF
--- a/nats-base-client/protocol.ts
+++ b/nats-base-client/protocol.ts
@@ -442,7 +442,7 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
   }
 
   processPing() {
-    this.transport.send(PONG_CMD);
+    this.transport.send(PONG_CMD).catch(() => {/*ignoring socket error*/});
   }
 
   processPong() {
@@ -667,7 +667,9 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
       }
     });
     if (cmds.length) {
-      this.transport.send(encode(cmds.join("")));
+      this.transport.send(encode(cmds.join(""))).catch(
+        () => {/*ignoring socket error*/},
+      );
     }
   }
 
@@ -722,7 +724,7 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
 
     if (this.outbound.size()) {
       const d = this.outbound.drain();
-      this.transport.send(d);
+      this.transport.send(d).catch(() => {/*ignoring socket error*/});
     }
   }
 


kozlovic (approved):
LGTM

intech:
Errors suppression was never considered a good pattern - wasn't it?
img

intech:
@aricart @kozlovic ping

aricart:
In this particular case it slipped through - the correct handling of the error was elsewhere but those promises were forgotten :(

aricart:
In case it is not clear, these events fail when the socket already failed, which is handled on the transport.

intech:
@aricart A more attractive solution would be to create an error event and emit such errors so that they can always be logged and processed if necessary.

aricart:
The disconnect is reported via the status mechanism.  If your app requires to know if your messages arrived, you use request reply.

aricart:
@arjenvdhave that one is in catch but not awaited, may need to modify the transport interface.

arjenvdhave:
Yep me to :)
For some reason I only get this in my app when it is running in a docker container on our k8s cluster.
I still haven't got a clue what the cause is.

Other services using the same setup don't have this issue.

aricart:
@arjenvdhave the network overlay is possibly the problem, right now trying to remove the async nature of send - as I am interested in a possible exception that happens as part of the write operation, but not the actual write succeeding... 
Do you have a repeatable case, if I publish a bundle for you to try?

arjenvdhave:
I hacked the `catch` call in my docker image.
I now get this error:

```
error: NatsError: DISCONNECT
- error.js:100 Function.errorForCode
[app]/[nats]/lib/nats-base-client/error.js:100:16
- protocol.js:175
[app]/[nats]/lib/nats-base-client/protocol.js:175:40
- Array.forEach
- protocol.js:174 ProtocolHandler.resetOutbound
[app]/[nats]/lib/nats-base-client/protocol.js:174:15
- protocol.js:195 ProtocolHandler.prepare
[app]/[nats]/lib/nats-base-client/protocol.js:195:14
- protocol.js:244 ProtocolHandler.<anonymous>
[app]/[nats]/lib/nats-base-client/protocol.js:244:31
- Generator.next
- protocol.js:34
[app]/[nats]/lib/nats-base-client/protocol.js:34:75
- new Promise
- protocol.js:13 __awaiter
[app]/[nats]/lib/nats-base-client/protocol.js:13:16
```

And after that it retries and connects without issues

arjenvdhave:
@aricart i can reproduce the issues on my cluster.
If you have "fix" i can try it.

I however do use the nats-js lib, not this deno directly

aricart:
give me a few, posting the branch for the deno side, now need to change the usage on nats.js

Issue #292, opened 2022-05-11
support subject report on StreamInfo subject report (ADR 91)


aricart:
https://github.com/nats-io/nats-architecture-and-design/issues/91

PR #293, opened 2022-05-11
[FEAT] support inspecting subjects in stream via StreamInfo - ADR issue 91

aricart:


aricart:
FIX #292

aricart:
FIX #292

tests/jsm_test.ts
```diff
@@ -984,3 +984,27 @@ Deno.test("jsm - update consumer", async () => {
...
+  const jsm = await nc.jetstreamManager();
+  const name = nuid.next();
+  await jsm.streams.add({ name, subjects: [`${name}.>`] });
+
+  const js = nc.jetstream();
+  await js.publish(`${name}.a`);
+  await js.publish(`${name}.a.b`);
+  await js.publish(`${name}.a.b.c`);
+
+  const si = await jsm.streams.info(name, { subjects_filter: `>` });
```
kozlovic, line 1002: I would add a test after that where you ask for info without the subjects_filter field and assert that there is no list returned. Also, I may have use a more restrictive filter such as `${name}.a.>` and expect only 2 subjects in `si.state.subjects`?

kozlovic (approved):
LGTM, but would expand on the test, at very least check that without filter field, si.state.subjects is not set.

tests/jsm_test.ts
aricart: done.

kozlovic (approved):
LGTM (I suggested more restrictive than ">" as a replacement, not addition, but that's fine).

PR #294, opened 2022-05-12
`conn.closed()` resolved error not always the actual reason for the error

### LINKED ISSUE #268: Authentication failures sometimes result in `ConnectionRefused` exceptions instead
When the reconnect attempts are exhausted, a `ConnectionRefused` exception is thrown.

https://github.com/nats-io/nats.deno/blob/afc6da4cea96acc4b5b8dfc225bf3289ad527db6/nats-base-client/protocol.ts#L342

However, this can occur even for expired authentication cases (it does in my application) and it's impossible to determine whether there is a problem connecting to the server (e.g.: it's down) or if some other thing happened. Probably a more specific error would help determine what caused the attempts to be exhausted.

aricart:
[FIX] while runtime authorization errors are notified, errors related to connection lastError were not properly reported by the closed() promise - this typically resulted in a "Connection Refused" rather than the actual reason.

FIX #268

### DIFF
--- a/nats-base-client/protocol.ts
+++ b/nats-base-client/protocol.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2021 The NATS Authors
+ * Copyright 2018-2022 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

PR #296, opened 2022-05-13
Expose permission errors at the request api level

### LINKED ISSUE #266: request error enhancement
If a client has pub permission error for the subject of a request, the request will simply come back as having timed out  - this is the wrong expedition for the developer - specially if they are not monitoring the status.

on publishing a request we should tag the associated subscription with the request subject, then on handle error we could test the subject to see if was the result of a request and provide a better experience.

aricart:
[FEAT] request api only rejects responses due to timeouts or no responders, but in the case of permissions, the client may not be able to publish to the request subject or subscribe to the mux inbox, this change rejects the request providing a more context relevant error. Previously it was only possible to learn about the error via the async error notification.

FIX #266

### DIFF
--- a/nats-base-client/error.ts
+++ b/nats-base-client/error.ts
@@ -96,6 +96,7 @@ export class NatsError extends Error {
   name: string;
   message: string;
   code: string;
+  permissionContext?: { operation: string; subject: string };
   chainedError?: Error;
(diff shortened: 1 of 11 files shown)


kozlovic (approved):
LGTM

PR #302, opened 2022-05-16
[FEAT] implements the ability in KV to request a specific revision for a key

### LINKED ISSUE #301: KV get revision
https://github.com/nats-io/nats-architecture-and-design/issues/93

aricart:
FIX #301

### DIFF
--- a/nats-base-client/kv.ts
+++ b/nats-base-client/kv.ts
@@ -312,10 +312,10 @@ export class Bucket implements KV, KvRemove {
     return data.length;
   }
 
-  smToEntry(key: string, sm: StoredMsg): KvEntry {
+  smToEntry(sm: StoredMsg): KvEntry {
     return {
       bucket: this.bucket,
-      key: key,
+      key: sm.subject.substring(this.prefixLen),
       value: sm.data,
       delta: 0,
       created: sm.time,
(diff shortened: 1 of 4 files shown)


nats-base-client/kv.ts
```diff
@@ -367,13 +368,20 @@ export class Bucket implements KV, KvRemove {
...
+    let arg: MsgRequest = { last_by_subj: this.fullKeyName(ek) };
+    if (opts.revision !== 0) {
+      arg = { seq: opts.revision };
+    }
+
     try {
-      const sm = await this.jsm.streams.getMessage(this.bucketName(), {
-        last_by_subj: this.fullKeyName(ek),
-      });
+      const sm = await this.jsm.streams.getMessage(this.bucketName(), arg);
```
kozlovic, line 384: There is a nuance in the Go client: when asking for a revision (sequence number), it is possible that a message is found at that sequence, but that does not mean that it will match the subject (the key). So if you had "A"/seq 1, "B"/seq 2 and you ask Get("A", 2), a message will be returned by the low level "getMessage(2)", but that does not match the key/subject "A", so you should return "key not found" in that case.

tests/kv_test.ts
```diff
@@ -1271,3 +1271,35 @@ Deno.test("kv - watch init callback exceptions terminate the iterator", async ()
...
+  }
+
+  await check("c");
+  await check("a", 1);
+  await check("b", 2);
+
+  await b.put("A", sc.encode("d"));
+  await check("d");
+  await check(null, 1);
+
```
kozlovic, line 1303: Should extend the test to demonstrate the issue I was referring to. Do a "put("B", someValue)" and then a "get("A", rev==5)", and you should have a failure (and not whatever value was stored in "B").

kozlovic (changes requested):

kozlovic (approved):
LGTM!

PR #306, opened 2022-05-17
[FIX] Include headers when sending a request with `noMux`

### LINKED ISSUE #305: Requests with a custom reply subject and headers do not forward headers
When doing a request that contains both a custom reply subject (thus `noMux: true`) will not send the headers with the request:

```typescript
const requestHeaders = headers();
requestHeaders.append("my-header", "Hello, World!");
const response = await this.connection.request(subject, encodedPayload, {
  reply: `_INBOX.something-custom`,
  noMux: true,
  headers: requestHeaders,
});
```

On the server side no headers will be received. That's because headers are not part of the options fowarded to the `publish` call: https://github.com/nats-io/nats.deno/blob/bad50bf2f6e1ac15a653b534a9b145a239ec3dea/nats-base-client/nats.ts#L202

This is unlike the case where `noMux: false`: https://github.com/nats-io/nats.deno/blob/bad50bf2f6e1ac15a653b534a9b145a239ec3dea/nats-base-client/nats.ts#L214

Kjir:
When sending a request with a custom subject for the reply, `noMux` must
be set to `true`. In this case, however, the eventual headers were not
forwarded to the server when performing the actual request. With this
commit, the headers are included as well in the `noMux` case.

FIX #305

### DIFF
--- a/nats-base-client/nats.ts
+++ b/nats-base-client/nats.ts
@@ -199,7 +199,10 @@ export class NatsConnectionImpl implements NatsConnection {
         },
       );
       (sub as SubscriptionImpl).requestSubject = subject;
-      this.protocol.publish(subject, data, { reply: inbox });
+      this.protocol.publish(subject, data, {
+        reply: inbox,
+        headers: opts.headers,
+      });
       return d;
     } else {
       const r = new Request(this.protocol.muxSubscriptions, subject, opts);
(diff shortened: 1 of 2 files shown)


aricart (approved):
LGTM

aricart:
@Kjir thank you!

Kjir:
Thank you for the quick response!

PR #307, opened 2022-05-17
misc client parity functionality

aricart:
[FEAT] nc#rtt() returns the number of millis for a round trip to the server
[FEAT] jsm#getAccountInfo() now reports tiered usage and limits
[TEST] added a test validating underlying jetstream error codes


kozlovic (approved):
LGTM

PR #311, opened 2022-05-20
[FEAT] [EXPERIMENTAL] requestMany()

aricart:
[FEAT] [EXPERIMENTAL] nats core requestMany() - allows implementing scatter gather receiving multi-message responses.

### DIFF
--- a/nats-base-client/request.ts
+++ b/nats-base-client/request.ts
@@ -12,34 +12,131 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Deferred, deferred, extend, Timeout, timeout } from "./util.ts";
-import type { Msg, RequestOptions } from "./types.ts";
+import { Deferred, deferred, Timeout, timeout } from "./util.ts";
+import type { Msg, RequestManyOptions, RequestOptions } from "./types.ts";
+import { RequestStrategy } from "./types.ts";
 import { ErrorCode, NatsError } from "./error.ts";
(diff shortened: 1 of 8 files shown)


nats-base-client/request.ts
```diff
@@ -12,35 +12,132 @@
...
+        if (this.max === 0) {
+          this.cancel();
+        }
+      }
+
+      if (this.opts.strategy === RequestStrategy.JitterTimer) {
+        clearTimeout(this.timer);
+        this.timer = setTimeout(() => {
+          this.cancel();
+        }, 300);
```
kozlovic, line 115: If I understand correctly, with this mode, and if no "max count" is specified (I can't tell if this is mandatory or not, based on test you added, looks like it is not), then possibly you would get an unbounded number of replies, right? Meaning that although one could set a max_wait of 2sec, if a message comes, we reset the timer to fire in 300ms here, but I don't see something that would make it stop because we are passed the original max 2 sec...

aricart: correct - this is the scatter pattern that Derek always talks about - I use it for ngs - Effectively it has an original deadline of some amount of time, but as messages come in that deadline is reset to a much smaller time. While the test was shrunk, I made do 100,000K+ messages, and I get all of them, possibly we'll want to expose that smaller timeout, but I figure that for now we are good.

kozlovic (approved):
LGTM

PR #312, opened 2022-05-25
[FIX] change transport.send() to not return a promise

aricart:
Another unhandled promise on a K8S environment during the connect/ping exchange an EPIPE was getting generated because of aggressive throttling on resource limits (100 millis of a CPU given to the process).

Verified the change with the help of @arjenvdhave who was kind enough to test and reproduce in his environment.

See https://github.com/nats-io/nats.deno/pull/289#issuecomment-1137372830


kozlovic (approved):
LGTM

PR #315, opened 2022-05-31
[FIX] [JS] change jsm.streams.update() to apply updates to existing config

aricart:
[FIX] jsm.streams.update() to retrieve the existing stream configuration and merging new options with other values to prevent unexpected changes to the stream configuration.


caleblloyd (approved):
LGTM

PR #316, opened 2022-06-01
[FEAT] [EXPERIMENTAL] direct get of messages for KV

aricart:
- [FEAT] implemented getDirectMessage() experimental feature
- [FEAT] [KV] experimental - KvOptions can now specify `allow_direct` which enables the creation of streams on servers that supported that enable the direct api which may provide better performance on gets for KV.
- [FIX] internal checkJsErrorCode() was passing a status instead of a description, creating error descriptions prefixed by the error code

### DIFF
--- a/nats-base-client/jsutil.ts
+++ b/nats-base-client/jsutil.ts
@@ -89,7 +89,7 @@ export function checkJsError(msg: Msg): NatsError | null {
   if (!h) {
     return null;
   }
-  return checkJsErrorCode(h.code, h.status);
+  return checkJsErrorCode(h.code, h.description);
 }
 
 export function checkJsErrorCode(
--- a/nats-base-client/kv.ts
+++ b/nats-base-client/kv.ts
@@ -51,8 +51,9 @@ import {
 import { millis, nanos } from "./jsutil.ts";
 import { QueuedIterator, QueuedIteratorImpl } from "./queued_iterator.ts";
 import { headers, MsgHdrs } from "./headers.ts";
-import { consumerOpts, deferred } from "./mod.ts";
+import { consumerOpts, deferred, ErrorCode } from "./mod.ts";
 import { compare, parseSemVer } from "./semver.ts";
+import { StreamAPIImpl } from "./jsmstream_api.ts";
 
 export function Base64KeyCodec(): KvCodec<string> {
   return {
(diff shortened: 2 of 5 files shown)


nats-base-client/kv.ts
```diff
@@ -231,6 +234,10 @@ export class Bucket implements KV, KvRemove {
     const discardNew = have ? compare(have, parseSemVer("2.7.2")) >= 0 : false;
     sc.discard = discardNew ? DiscardPolicy.New : DiscardPolicy.Old;
 
+    const direct = have ? compare(have, parseSemVer("2.8.5")) >= 0 : false;
```
piotrpio, line 237: Seeing that we already have to check server version compatibility twice here, maybe it would be good to add and export a `checkCompat(have, want)` function in `semver.ts`?

nats-base-client/types.ts
```diff
@@ -943,6 +948,7 @@ export interface KvOptions extends KvLimits {
   streamName: string;
   codec: KvCodecs;
   bindOnly: boolean;
+  allow_direct: boolean;
```
derekcollison, line 951: This is for all streams, although agree it will be more prevalent with KV.

nats-base-client/types.ts
```diff
@@ -516,6 +516,11 @@ export interface StoredMsg {
   time: Date;
 }
 
+export interface DirectMsg extends StoredMsg {
+  stream: string;
```
derekcollison, line 520: This public? Should just look like a message you can get today with kv.Get()

nats-base-client/jsutil.ts
```diff
@@ -89,7 +89,7 @@ export function checkJsError(msg: Msg): NatsError | null {
   if (!h) {
     return null;
   }
-  return checkJsErrorCode(h.code, h.status);
+  return checkJsErrorCode(h.code, h.description);
```
derekcollison, line 92: We want to check description vs status?

nats-base-client/types.ts
aricart: 👍

nats-base-client/types.ts
aricart: No, it is not, they see exactly the same API

nats-base-client/jsutil.ts
aricart: the status line contains the code/message - the description is just the message part of that combo, so it uglified some of the errors.

piotrpio (approved):
LGTM

PR #318, opened 2022-06-03
[EXPERIMENTAL] [JETSTREAM] pull consumers can now specify `max_bytes` on `PullOptions`

aricart:
The server will limit the number of messages it sends to the `max_bytes` option. Note that if the stream contains a message with a payload larger than `max_bytes`, the consumer will fail. This means that the iterator will end with an error and callbacks will raise this error.

### DIFF
--- a/nats-base-client/jsclient.ts
+++ b/nats-base-client/jsclient.ts
@@ -70,6 +70,7 @@ import { headers } from "./headers.ts";
 import { consumerOpts, isConsumerOptsBuilder } from "./jsconsumeropts.ts";
 import { Bucket } from "./kv.ts";
 import { NatsConnectionImpl } from "./nats.ts";
+import { Feature } from "./semver.ts";
 
 export interface JetStreamSubscriptionInfoable {
   info: JetStreamSubscriptionInfo | null;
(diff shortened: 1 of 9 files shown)


nats-base-client/jsclient.ts
```diff
@@ -252,12 +262,18 @@ export class JetStreamClientImpl extends BaseApiClient
...
-            (err.code === ErrorCode.JetStream404NoMessages ||
-              err.code === ErrorCode.JetStream408RequestTimeout)
-          ) {
-            qi.stop();
+          if (isNatsError(err)) {
+            switch (err.code) {
+              case ErrorCode.JetStream404NoMessages:
+              case ErrorCode.JetStream409:
+                qi.stop();
+                return [null, null];
```
piotrpio, line 270: Does the user get no feedback on what happened if there is 409 status returned? Also, why do we return in this specific scenario?

piotrpio:
Shouldn't `fetch()` check for server compatibility before attempting to track bytes?

aricart:
> Shouldn't `fetch()` check for server compatibility before attempting to track bytes?

Yes it possibly should - currently, there's no version check at the client level, but perhaps it is time for this

aricart:
> Shouldn't `fetch()` check for server compatibility before attempting to track bytes?

Absolutely.

nats-base-client/jsclient.ts
piotrpio: Got it, make sense. I assumed 404 means there are no more messages.

piotrpio (approved):
LGTM

PR #327, opened 2022-07-07
consumer heartbeats

aricart:
[FEAT] added optional heartbeat checks to js#fetch()
[FEAT] added optional heartbeat checks to js#subcribe()


kozlovic (approved):
LGTM

PR #330, opened 2022-07-13
[FIX] name of the consumer was not captured when binding to an ephemeral

### LINKED ISSUE #329: sub.consumerInfo() doesn't work for ephemerals
The name of the ephemeral is not captured correctly in the subscription if binding

aricart:
This has the effect of failing `consumerInfo()` on the subscription

FIX #329


kozlovic (approved):
LGTM

PR #337, opened 2022-07-27
[FEAT] Consumer support for specifying num_replicas

aricart:


nats-base-client/types.ts
```diff
@@ -2392,7 +2397,8 @@ export interface ConsumerUpdateConfig {
    */
   "max_bytes"?: number;
   /**
-   * When set do not inherit the replica count from the stream but specifically set it to this amount
+   * When set do not inherit the replica count from the stream but specifically set it to this amount.
+   * Min is 1, Max is 5
```
kozlovic, line 2401: I don't think library should enforce a max, and not sure even server does at this time (there may be configuration to enforce a max, but that should be left to the server/config and not at the library level)

kozlovic (approved):
LGTM, but would remove the note about replicas max value.

nats-base-client/types.ts
aricart: done.

kozlovic (approved):
LGTM

PR #341, opened 2022-08-04
Direct API get last_by_subject enhancement

aricart:
[FEAT] Direct API get for last message by subject will now set the subject as part of the request subject, rather than include it on the payload of the request. This allows for downstream clients, such as KV, to clamp on permissions for a GET.


kozlovic (approved):
LGTM

PR #347, opened 2022-08-15
fix: support optional tls connections if available with the server

reecefenwick:
- Fixes and issue where TLS was only allowed if required by the server, if tls is not required clients should still be able to connect

**Issue**
When connecting to a NATS server that supports TLS but does not require it the `nats.deno` client cannot connect to it.

**Fix**
Throw error when validating TLS config only when `options.tls` is provided AND `ServerInfo.tls_available` is false

### DIFF
--- a/nats-base-client/options.ts
+++ b/nats-base-client/options.ts
@@ -128,11 +128,11 @@ export function parseOptions(opts?: ConnectionOptions): ConnectionOptions {
 }
 
 export function checkOptions(info: ServerInfo, options: ConnectionOptions) {
-  const { proto, tls_required: tlsRequired } = info;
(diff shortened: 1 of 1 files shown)


aricart:
@reecefenwick we also need it to trigger if TLS is required. Otherwise the client won't connect.

aricart:
@reecefenwick The change is a bit more complex than shown here, right now the option is a patch for rejection if the server supports it, but the handling of the upgrade is not implemented in the PR. Mind if I take it over?

aricart:
Closing this PR in favor of https://github.com/nats-io/nats.deno/pull/348. Thanks @reecefenwick.

PR #348, opened 2022-08-15
[FIX] support tls_available

aricart:
servers that enable `allow_non_tls` report that they have `tls_available`. This fix enables auto-upgrade by the client to a tls connection unless the `tls` connect option is set to `null`. Previously the client was only responding to `tls_required`.


aricart:
This PR addresses the actual tls upgrade.
https://github.com/nats-io/nats.deno/pull/347

Thanks to @reecefenwick

kozlovic (approved):
LGTM

PR #349, opened 2022-08-22
[FIX] kv status is exposed to the client as millis on set, but on status() it was returned raw in nanos

### LINKED ISSUE #317: KV status reports the ttl in nanos, but on creation, the limit is specified in millis
the api needs to be transitive, for this operation it is expected to get millis back

aricart:
FIX #317


bruth (approved):
LGTM

PR #358, opened 2022-09-12
[FIX] fixed an issue where the correct info for an object entry was not calculated

### LINKED ISSUE #357: New object store file upload example
When using NodeJS to upload a file via like:
```js
await os.put({name: 'filename.gz'}, this.readableStreamFrom(fs.readFileSync(./filename.gz)))
```

```js
readableStreamFrom(data: Uint8Array): ReadableStream<Uint8Array> {
  return new ReadableStream<Uint8Array>({
    pull(controller) {
      controller.enqueue(data);
      controller.close();
    }
  })
}
```
The file uploads properly but when downloading the file (using nats cli) it is corrupt and wont extract. It shouldn't matter but the example uses a gzipped file.

Any pointers for how the file should be passed in? The example is just encoding/decoding a string but I believe files are read in as a buffer type.

Thanks.

aricart:
[FIX] fixed an issue where the correct info for an object entry was not calculated when the entry was empty or didn't have pending chunk changes that were unprocessed.

[FIX] made the underlying stream info requests trap the stream not found and instead return "object store not found" [TEST] added more tests to object store

FIX #357

### DIFF
--- a/nats-base-client/objectstore.ts
+++ b/nats-base-client/objectstore.ts
@@ -260,8 +260,23 @@ export class ObjectStoreImpl implements ObjectStore {
     }
   }
 
+  async _si(
+    opts?: Partial<StreamInfoRequestOptions>,
+  ): Promise<StreamInfo | null> {
+    try {
+      const info = await this.jsm.streams.info(this.stream, opts);
+      return info;
+    } catch (err) {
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

PR #363, opened 2022-09-15
[FIX] [OS] added additional checks to update

aricart:
- updating meta of a deleted object is not allowed
  - updating the meta to point an already existing name is not allowed


piotrpio (approved):
LGTM

PR #364, opened 2022-09-15
list kvs and objectstores

aricart:
[FEAT] added `jsm.streams.listKvs()` to provide a list of all the `KvStatus` for all streams that use the kv naming strategy.

[FEAT] [BETA] added `jsm.streams.listObjectStores()` to provide a list of all the `ObjectStoreStatus` for all streams that use the objectstore naming strategy.

[FEAT] [KV] `KvStatus` now has the property `streamInfo` with the `StreamInfo` for the KV.

[CHANGE] [BREAKING] [KV] `KvStatus.bucket` is now the user-specified name of the KV used when created - not the stream backing it. This is consistent with the Go client. Previously it returned `KV_` + name. For the actual stream name see `KvStatus.streamInfo.config.name`

[CHANGE] `ObjectStoreInfo` is deprecated use `ObjectStoreStatus` instead.

### DIFF
--- a/nats-base-client/jsmstream_api.ts
+++ b/nats-base-client/jsmstream_api.ts
@@ -16,10 +16,12 @@
 import {
   Empty,
   JetStreamOptions,
+  KvStatus,
   Lister,
   MsgDeleteRequest,
   MsgRequest,
   NatsConnection,
+  ObjectStoreStatus,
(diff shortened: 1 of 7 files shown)


kozlovic (approved):
LGTM

PR #365, opened 2022-09-16
StreamInfo pagination support

aricart:
[FEAT] added internal pagination to `jsm.streams.info()`, if required, the call will collect the necessary pages and merge `subjects`. `StreamInfoRequestOptions` now also has support for `ApiPagedRequest`.

[FIX] `Lister` interface was missing `Symbol.asyncIterator`, so it couldn't be used directly as intended.


kozlovic (approved):
LGTM

PR #366, opened 2022-09-16
Feature disablements

aricart:
[FEAT] added mechanism where a `Feature` can be disabled by the user. This requires casting the connection to a `NatsConnectionImpl`, and then accessing the `Features` controller as `nc.features.disable(Feature)`. This is an experimental feature, currently supporting a very small number of options


kozlovic (approved):
LGTM

PR #367, opened 2022-09-19
[FIX] [OS] made `put()` reject link configurations

aricart:
Link configurations must be routed through the `link()` API. This matches changes made to the go client in https://github.com/nats-io/nats.go/pull/1057/files


kozlovic (approved):
LGTM

PR #369, opened 2022-09-19
[FEAT] `jsm.streams.list()` can now specify a subject filter

aricart:
This will limit the list to include streams that include the specified subject.


kozlovic (approved):
LGTM

PR #372, opened 2022-09-21
[FIX] cross language compatible meta subject

tothandras:
I'm opening this PR mainly to open a discussion around language cross compatibility (with Go). This would be a breaking change.
https://github.com/nats-io/nats.go/blob/f4102422df7df83b2032318fd14a248c45f113e1/object.go#L442

also related: https://github.com/nats-io/nats.py/pull/331#pullrequestreview-1116024414

### DIFF
--- a/nats-base-client/base64.ts
+++ b/nats-base-client/base64.ts
@@ -1,0 +1,41 @@
+export class Base64Codec {
+  static encode(bytes: string | Uint8Array): string {
+    if (typeof bytes === "string") {
+      return btoa(bytes);
+    }
+    const a = Array.from(bytes);
+    return btoa(String.fromCharCode(...a));
+  }
+
+  static decode(s: string, binary = false): Uint8Array | string {
+    const bin = atob(s);
+    if (!binary) {
+      return bin;
+    }
+    return Uint8Array.from(bin, (c) => c.charCodeAt(0));
+  }
+}
+
+export class Base64UrlCodec {
+  static encode(bytes: string | Uint8Array): string {
+    return Base64UrlCodec.toB64URLEncoding(Base64Codec.encode(bytes));
+  }
+
+  static decode(s: string, binary = false): Uint8Array | string {
+    return Base64Codec.decode(Base64UrlCodec.fromB64URLEncoding(s), binary);
+  }
+
+  static toB64URLEncoding(b64str: string): string {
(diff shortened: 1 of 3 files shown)


aricart:
Yes, this is a bug in the javascript client, the proposed fix doesn't work because Buffer is only available in the node.js ecosystem.

tothandras:
oh, indeed! https://deno.land/std/encoding/base64url.ts should work, I'll replace the Buffer and try it out as well.

aricart:
the solution has to work on a browser where std/encoding is not available.

aricart:
```
class Base64Codec {
  static encode(bytes: string | Uint8Array): string {
    if (typeof bytes === "string") {
      return btoa(bytes);
    }
    const a = Array.from(bytes);
    return btoa(String.fromCharCode(...a));
  }

  static decode(s: string, binary = false): Uint8Array | string {
    const bin = atob(s);
    if (!binary) {
      return bin;
    }
    const bytes = new Uint8Array(bin.length);
    for (let i = 0; i < bin.length; i++) {
      bytes[i] = bin.charCodeAt(i);
    }
    return bytes;
  }
}

class Base64UrlCodec {
  static encode(bytes: string | Uint8Array): string {
    return Base64UrlCodec.toB64URLEncoding(Base64Codec.encode(bytes));
  }

  static decode(s: string, binary = false): Uint8Array | string {
    return Base64Codec.decode(Base64UrlCodec.fromB64URLEncoding(s), binary);
  }

  static toB64URLEncoding(b64str: string): string {
    b64str = b64str.replace(/=/g, "");
    b64str = b64str.replace(/\+/g, "-");
    return b64str.replace(/\//g, "_");
  }

  static fromB64URLEncoding(b64str: string): string {
    // pads are % 4, but not necessary on decoding
    b64str = b64str.replace(/_/g, "/");
    b64str = b64str.replace(/-/g, "+");
    return b64str;
  }
}
```

aricart:
I'll put the above somewhere conventient

tothandras:
@aricart Something like this https://github.com/nats-io/nats.deno/pull/372/commits/b4cd346b6244443023e4b2c5216149025517e5e3?

aricart (approved):
LGTM

aricart:
@tothandras thanks for the contribution!

PR #374, opened 2022-09-28
[FIX] [TEST] cross account test environment to allow acks to cross

aricart:


tests/jetstream_test.ts
```diff
@@ -1496,9 +1496,11 @@ Deno.test("jetstream - cross account subscribe", async () => {
       m.ack();
     }
   })();
+  await nc.flush();
```
kozlovic, line 1499: More importantly, if you have a "sync" version of m.ack() (that is, the one that waits on a reply for the ACK) that would be better. The server is handling ACKs in a go routine so it is very possible that after a non-sync ack() the message is still not acked in the server. Doing nc.flush() will likely "work", but not guaranteed. Alternatively, you may repeat the check below (consumer info/ackfloor, etc...) for a bit...

kozlovic (approved):
LGTM

tests/jetstream_test.ts
aricart: yea - I ended up doing something else - in the above, the sub should end after 2 messages, but the issue is the outbound may have not yet been sent. But this possibly prevents a flapper.

kozlovic (approved):
LGTM

PR #379, opened 2022-09-29
[FEAT] [JS] added JSM API to list stream names

aricart:
An optional subject can be specified to filter to include only streams that contain or match the specified subject (can contain wildcards).


kozlovic (approved):
LGTM

PR #388, opened 2022-10-24
subscription closed, terminated iterators that had pending messages

aricart:
[FIX] fixed an issue where if an unsubscribe was set on a subscription and the subscription used an iterator, the iterator was stopped immediately rather than waiting for the currently in-flight messages to be processed.


kozlovic (approved):
LGTM

PR #390, opened 2022-10-25
Early PONG rejection

aricart:
[FIX] added a catch to the initial pong promise as it seems possible in some difficult scenarios in node for the pong to be rejected before other handlers are setup. See https://github.com/nats-io/nats.js/issues/523

[LINT] fixed some linter and fmt warnings

### DIFF
--- a/README.md
+++ b/README.md
@@ -838,4 +838,6 @@ The relationship between these is:
 
 ## JetStream
 
-JetStream is the NATS persistence engine providing streaming, message, and worker queues with At-Least-Once semantics. [Support for JetStream is built-in](jetstream.md).
+JetStream is the NATS persistence engine providing streaming, message, and
+worker queues with At-Least-Once semantics.
+[Support for JetStream is built-in](jetstream.md).
--- a/nats-base-client/protocol.ts
+++ b/nats-base-client/protocol.ts
@@ -217,6 +217,9 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
     this.resetOutbound();
 
     const pong = deferred<void>();
+    pong.catch(() => {
+      // provide at least one catch - as pong rejection can happen before it is expected
+    });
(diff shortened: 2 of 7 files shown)


kozlovic (approved):
LGTM

nats-base-client/protocol.ts
```diff
@@ -217,6 +217,9 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
     this.resetOutbound();
 
     const pong = deferred<void>();
+    pong.catch(() => {
+      // provide at least one catch - as pong rejection can happen before it is expected
```
ArmorDarks, line 221: How about a log warning about such errors, or better make it properly awaited? There's no guarantee that it will always fail with an "ok" error. If some exception happens, it will be swallowed here, which will make debugging quite a challenge

aricart:
Added some additional guardrails, depending on the runtime the type of error returned will be different. But in cases where the client was telegraphing the PONG rejection, the disconnect error will not have a trace, as that trace is bogus, the client simply disconnected. With that said in node, the errors are properly reported now, but that is a different PR.

README.md
```diff
@@ -838,4 +838,6 @@ The relationship between these is:
 
 ## JetStream
 
-JetStream is the NATS persistence engine providing streaming, message, and worker queues with At-Least-Once semantics. [Support for JetStream is built-in](jetstream.md).
+JetStream is the NATS persistence engine providing streaming, message, and
```
kozlovic, line 841: I have seen this change in an older PR, no? Could it be that your branch was out-of-date? Maybe a rebase would help?

kozlovic (approved):
LGTM

README.md
aricart: the nats.js has the same comment - the current version of deno reformatted that file-

PR #391, opened 2022-10-27
KV mirror

aricart:


nats-base-client/jsmstream_api.ts
```diff
@@ -83,6 +109,9 @@ export class StreamAPIImpl extends BaseApiClient implements StreamAPI {
     validateStreamName(name);
     const old = await this.info(name);
     const update = Object.assign(old.config, cfg);
+    update.mirror = convertStreamSourceDomain(update.mirror);
```
kozlovic, line 112: Not sure updating a stream and changing the mirror/source actually works (in that I don't think the server supports that), so maybe the reason why it is only on "AddStream" in the go client?

aricart: it should be a noop regardless - because it has to have "domain" set which won't be, so it is ignored. I just wanted to trap it for uniformity in case it updates later.

kozlovic (approved):
LGTM

aricart:
@kozlovic I added the additional changes required by the client to support history, watch, keys.

PR #392, opened 2022-11-02
[CI] updated nats-server to 2.9.5, and deno to 1.27.0

aricart:


kozlovic (approved):
LGTM. You have a test (about rtt?) that fails on CI but I think you fixed a flapper on a recently merged PR, so that may be that one.

PR #397, opened 2022-11-03
[SERVICE] added "ServiceError" which allows the client to set a code message which is serialized into a service header

aricart:


piotrpio (approved):
LGTM

PR #409, opened 2022-11-17
[FEAT] added connect option `ignoreAuthErrorAbort` which disables a connection abort if client fails to authenticate twice with the same error.

aricart:


derekcollison (approved):
LGTM

PR #410, opened 2022-11-17
Merging two fixes from dev that remove ambiguity on connection errors

aricart:
These were already reviewed in dev, but will benefit the setup having connection problems behind kubernetes LB.


kozlovic (approved):
LGTM

PR #412, opened 2022-11-18
Authenticator Enhancements

aricart:
[FEAT] authenticator functions now delay executing user functions until required. This means that user functions can provide different authentication values when the authentication happens. Because these functions execute during the client handshake, they must not take significant time or the server may timeout the authentication process. The recommended strategy is for the client to asynchronously maintain up-to-date credentials, that can be presented when required.

[FEAT] All authenticators can now provide functions for all their arguments.


kozlovic (approved):
LGTM

derekcollison (approved):
LGTM

PR #413, opened 2022-11-18
[FIX] processInfo random error was mapped as BadAuthentication, when it should just be whatever error it was

aricart:
if during processInfo there was some sort of exception (most likely it is a bad authenticator), the error was transformed to a BadAuthentication, this was incorrect and instead what should happen is the error is telegraphed as it is to the client.


derekcollison (approved):
LGTM

PR #420, opened 2022-11-30
[internal] pushing a funtion to a queued iterator that sets an error is ignored until the next yield

aricart:
[FIX] if internal code inserted an error into an iterator by pushing a function that set the error condition, the iterator wouldn't immediately yield the error, as the error check happened prior to the execution of the function

[TEST] added tests for idle heartbeat conditions.


piotrpio (approved):
LGTM!

PR #422, opened 2022-12-01
[FIX] protocol errors shoudn't close() the client

### LINKED ISSUE #421: Closed connection with 'Stale Connection' error doesn't auto-reconnect
* Client version: 2.8.0
* Node version: 16.14.2

## What's happening

Given the config

```ts
const connection = await connect({
  maxReconnectAttempts: -1,
  name: 'some-name',
  pass: 'some-pass',
  servers: ['load-balancer-url'],
  user: 'some-user',
  inboxPrefix: 'some.inbox',
})
```

If the connection closed with the following error

```
NatsError: 'Stale Connection'
    at Function.toError (/code/node_modules/nats/lib/nats-base-client/protocol.js:339:20)
    at ProtocolHandler.<anonymous> (/code/node_modules/nats/lib/nats-base-client/protocol.js:363:41)
    at Generator.next (<anonymous>)
    at /code/node_modules/nats/lib/nats-base-client/protocol.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/code/node_modules/nats/lib/nats-base-client/protocol.js:4:12)
    at ProtocolHandler.processError (/code/node_modules/nats/lib/nats-base-client/protocol.js:361:16)
    at ProtocolHandler.push (/code/node_modules/nats/lib/nats-base-client/protocol.js:461:22)
    at Parser.parse (/code/node_modules/nats/lib/nats-base-client/parser.js:307:45)
    at ProtocolHandler.<anonymous> (/code/node_modules/nats/lib/nats-base-client/protocol.js:197:45)
```

it will never reconnect despite `maxReconnectAttempts` set to `-1`

## What's expected

If `maxReconnectAttempts` set to `-1`, the connection should automatically try reconnecting. Otherwise, it should be highlighted in the docs that this particular case has to be handled manually.

aricart:
[FIX] protocol errors (any unknown error from the server) was handled by the client with a `close()`. Now while the connection could reset, reconnects are honored.

FIX #421

### DIFF
--- a/nats-base-client/protocol.ts
+++ b/nats-base-client/protocol.ts
@@ -234,7 +234,9 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
       .then(async (_err?) => {
         this.connected = false;
         if (!this.isClosed()) {
-          await this.disconnected(this.transport.closeError);
+          // if the transport gave an error use that, otherwise
+          // we may have received a protocol error
+          await this.disconnected(this.transport.closeError || this.lastError);
           return;
         }
       });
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #424, opened 2022-12-02
[CHANGE] SERVICE API changes

aricart:
- [CHANGE] separated service error header into two headers, one with the code the other with the message
- [CHANGE] clamped down on the service name to be a filesystem safe name 
- [CHANGE] STATUS is now STATS to better reflect its intention, things named with status, changed to be stats. 
- [CHANGE] PING now returns the name and id of the service and is intended for client RTT calculations - [CHANGE] control subject (for invoking monitoring) can now specify a prefix (this replaces `$SRV` and is intended for cross account).


piotrpio (approved):
LGTM

PR #427, opened 2022-12-08
JetStream Simplification

aricart:


nats-base-client/consumermessages.ts
```diff
@@ -0,0 +1,449 @@
...
+          if (isHeartbeatMsg(msg)) {
+            return;
+          }
+          const code = msg.headers?.code;
+          const description = msg.headers?.description?.toLowerCase() ||
+            "unknown";
+          const { msgsLeft, bytesLeft } = this.parseDiscard(msg.headers);
+          if (msgsLeft > 0 || bytesLeft > 0) {
+            this.pending.msgs -= msgsLeft;
+            this.pending.bytes -= msgsLeft;
```
piotrpio, line 126: I think this should be:
```
this.pending.bytes -= bytesLeft;
```

nats-base-client/consumermessages.ts
```diff
@@ -0,0 +1,449 @@
...
+      `${this.consumer.api.prefix}.CONSUMER.MSG.NEXT.${this.consumer.stream}.${this.consumer.name}`,
+      this.consumer.api.jc.encode(opts),
+      { reply: this.inbox },
+    );
+    this.pending.bytes += opts.max_bytes ?? 0;
+    this.pending.msgs += opts.batch ?? 0;
+    this.pending.requests++;
+  }
+
+  pullOptions(): Partial<PullOptions> {
```
piotrpio, line 304: I wonder whether we should not send same request payloads for each request, e.g. for `max_messages == 100` and `threshold == 0.25` always send `batch: 25`. We will refill and not overshoot either way but (especially with bytes) it becomes more predictable.

aricart: but that assumes even message sizes - in the case where the payloads are not the same, it likely would stall the client or am I misunderstanding.

piotrpio: But your solution does not really prevent stalling if I understand correctly - it just sets `max_bytes` in pull request between the 25% - 100% of `max_bytes` option set by the user (assuming threshold is 0.75). Plus we don't have control over it, so it is very much possible that you send a request with 25% `max_bytes` option and the next message stalls either way... Only in that case we are never sure what the value in payload will be, so the most we can say for sure is "messages up to 25% of `max_bytes` will fit for sure, larger ones maybe will and maybe won't", which is not much of a guarantee.

By having constant `max_bytes` we can clearly document the behavior. Unless I'm misunderstanding the code or how `max_bytes` are handled by server.

aricart: The issue for the stall is if the messages don't fit the constraints of max_bytes - at that point, the server will send a response saying that it couldn't do it, and then the client will send a new request, this time presumably for larger max_bytes as it has processed additional messages - that will continue until the client can make progress. If the request is constant, it is possible it will never be able to make progress.

aricart: Lets chat about it

piotrpio (approved):
LGTM

PR #429, opened 2022-12-12
[CHANGE] [experimental] mrequest iterator signature

aricart:
[CHANGE] mrequest iterator signature - possible errors at core NATS level can only be sub/no responder type errors which should fail the request


scottf (approved):
LGTM

PR #434, opened 2022-12-16
[UPDATE] deno to 1.29.1 and std to 0.168.0

aricart:
[FIX] utility test functions that relied on older apis
[FIX] linter errors revealed by new version of deno


piotrpio (approved):
LGTM

PR #439, opened 2022-12-19
cross-client service verification tool

aricart:
cross client service metadata verification
requires a service:
- that echos back payload
- returns a service error message if payload is empty
- throws an error (or registers an internal error) if the payload value is "error"

clients can execute as:
`deno run -A https://raw.githubusercontent.com/nats-io/nats.deno/main/tests/helpers/service_metadata.ts --server demo.nats.io --name serviceName`


wallyqs (approved):
LGTM

PR #441, opened 2022-12-20
[CHANGE] [SRV] `name` and `id` are used in whatever case they are specified - this means that for monitoring and discovery, names/ids are significant

aricart:


piotrpio (approved):
LGTM

PR #443, opened 2022-12-21
[BUMP] nkeys.js dependency to 1.0.4

aricart:
[BUMP] nkeys.js dependency to 1.0.4 as downstream projects (nats.ws) when in a cjs module configuration with some bundlers would failed to propagate the dependency

[BUMP] client version to 1.10.2


piotrpio (approved):
LGTM

PR #446, opened 2022-12-23
[DOC] [JS] heatbeat example was not up-to-date, updated example showing behaviour of missed heartbeats on JS push subscriber.

aricart:


scottf (approved):
LGTM

PR #449, opened 2022-12-29
[FEAT] [SRV] response types

aricart:


nats-base-client/service.ts
```diff
@@ -50,7 +50,21 @@ export enum ServiceVerb {
   SCHEMA = "SCHEMA",
 }
 
-export interface ServiceIdentity {
+export enum ServiceResponseType {
+  STATS = "io.nats.micro.v1.stats",
```
ripienaar, line 54: https://github.com/nats-io/nats.go/blob/07b2b8f75b591b04543abd776983e29735aeef91/micro/service.go#L192

ripienaar (approved):
LGTM

PR #450, opened 2022-12-29
[FIX] prevent client from closing due to expired credentials if there's a subsequent reconnect after an expired credential

aricart:
This type of recovery can only happen if the client implements a jwt/creds handler


derekcollison (approved):
LGTM

PR #453, opened 2023-01-04
Allow # in stream name validation function

idanasulinmemphis:
I faced an issue when I tried to pullSubscribe from a stream whose name contains a # character.
I got the following error 
Invalid stream name - stream name cannot contain '#'.

So I followed up and saw that in your last releases, you changed something in the validation function.
Since # char is allowed according to your docs and I also succeeded to create a stream with #, I've changed a bit the regex you are using in order to validate the stream name

### DIFF
--- a/nats-base-client/jsutil.ts
+++ b/nats-base-client/jsutil.ts
@@ -49,7 +49,7 @@ export function validName(name = ""): string {
   if (name === "") {
     throw Error(`name required`);
   }
-  const RE = /^[-\w]+$/g;
+  const RE = /^[-\w,#]+$/g;
   const m = name.match(RE);
   if (m === null) {
     for (const c of name.split("")) {


aricart:
@idanasulinmemphis this was considered a bug on the javascript client. The allowed names for the stream are https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-6.md.  The server is not rejecting the name, but we are trying to avoid issues with names as these translate info filenames that run in a variety of filesystems.

aricart:
The interesting part is that the above list is not correct :(

idanasulinmemphis:
Sorry, but I didn't understand your offer.
I'm able to create a stream with # inside the name, so why can't I pull subscribe from it?
Also, it didn't fail on the broker side it fails on the SDK side.
BTW it is not happening in GO SDK for example

idanasulinmemphis:
any response?

aricart:
The fix is to set it back to what it was, and let the server decide to add more validations - this is only true for streams/consumers, but if in the context of KV/ObjectStore/service API, additional constraints will be required. I'll release later this week.

https://github.com/nats-io/nats.deno/pull/474

aricart:
https://github.com/nats-io/nats.deno/pull/474 implements the correct change

PR #456, opened 2023-01-11
srv schemas

aricart:
- [CHANGE] [SRV] `addEndpoint(name: string|{name: string, subject: string}, handler: ServiceHandler)` is now `addEndpoint(name: string, opts: ServiceHandler|EndpointOptions)`

  `EndpointOptions is {subject?: string, handler?:EndpointHandler, schema?: SchemaInfo}` - the previous signature was never released.

 - [FEAT] [SRV] now has a config option `apiURL: string` which can be an URL describing the documentation of the service
 
 - [CHANGE] [SRV] config's `schema` is now part of `endpoint`
 
 - [CHANGE] [SRV] ServiceSchema exposes `apiURL` and `endpoints` (`{name: string, subject: string, schema?: SchemaInfo}[]`)


piotrpio (approved):
LGTM

PR #457, opened 2023-01-11
[FEAT] [OS] expose timeout on put, for really large assets, it is possible for the publish request to timeout

aricart:
FIX https://github.com/nats-io/nats.ws/issues/182

### DIFF
--- a/nats-base-client/types.ts
+++ b/nats-base-client/types.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 The NATS Authors
+ * Copyright 2020-2023 The NATS Authors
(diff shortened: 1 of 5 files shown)


nats-base-client/types.ts
```diff
@@ -2919,13 +2919,21 @@ export type ObjectResult = {
   error: Promise<Error | null>;
 };
 
+export type ObjectStorePutOpts = {
```
piotrpio, line 2922: Update licence while you're here ;)

piotrpio (approved):
LGTM, just one suggestion

PR #469, opened 2023-01-19
[FEAT] multiple consumer filters

aricart:
[FEAT] consumer filter_subjects support - this feature is exclusive of the filter_subject but allows a consumer to filter multiple subjects in a stream. To set, specify filter_subjects instead of filter_subject or if using consumer options builder filterSubject() can be called multiple times.

[CHANGE] filter_subject (and filter_subjects) can be updated, the server removed this restriction


Jarema (approved):
LGTM

PR #471, opened 2023-01-23
[FIX] add support `previousRevision` in ObjectStorePutOps

aricart:
Currently, it is possible to add the same object twice, but not trigger the prune functionality, thus data assets are found twice in the stream. The option `previousRevision` allows the client to specify the revision that the previous object represented (`0` for not expecting the object to exist in the stream). If the update of the metadata entry fails, the data chunks are removed.


piotrpio (approved):
LGTM

piotrpio (approved):

PR #472, opened 2023-01-24
[FEAT] multiple authenticators

aricart:
- it is now possible to specify multiple authenticators, previously this was rejected by the client.
- all generated/specified authenticators are aggregated, and their authentication results are aggregated. Note that creds and jwt authenticators write the `nkey` and `sig` properties, so if both are specified, it is important that they both use the same nkey.


derekcollison (approved):
LGTM - caveats though, not TS person.

piotrpio (approved):
LGTM!

PR #477, opened 2023-02-06
[FEAT] Added json()/string() utilities to Msg and other variants

### LINKED ISSUE #476: [FEAT] json()/string() helpers on Msg
this would greatly simplify user code when reading messages, and still provide accountability for messages that are not parsable as the client expected.

aricart:
[FEAT] Added json()/string() utilities to Msg, JsMsg, KvEntry, StoredMsg, DirectMsg this allows clients to decode payloads easier while still requiring accountability on the client (as decoding can fail)

Fix #476

### DIFF
--- a/nats-base-client/kv.ts
+++ b/nats-base-client/kv.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021-2022 The NATS Authors
+ * Copyright 2021-2023 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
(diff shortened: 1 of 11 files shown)


nats-base-client/kv.ts
```diff
@@ -920,3 +902,103 @@ export class KvStatusImpl implements KvStatus {
...
+  get delta(): number {
+    return this.sm.info.pending;
+  }
+
+  get length(): number {
+    const slen = this.sm.headers?.get(JsHeaders.MessageSizeHdr) || "";
+    if (slen !== "") {
+      return parseInt(slen, 10);
+    }
+    return this.sm.data.length;
```
piotrpio, line 994: Shouldn't length be calculated as payload + headers?

nats-base-client/kv.ts
```diff
@@ -920,3 +902,103 @@ export class KvStatusImpl implements KvStatus {
...
+  get operation(): OperationType {
+    return this.sm.header.get(kvOperationHdr) as OperationType || "PUT";
+  }
+
+  get length(): number {
+    const slen = this.sm.header.get(JsHeaders.MessageSizeHdr) || "";
+    if (slen !== "") {
+      return parseInt(slen, 10);
+    }
+    return this.sm.data.length;
```
piotrpio, line 946: Shouldn't length be calculated as payload + headers?

nats-base-client/kv.ts
aricart: There are no headers in KV! - the only time the header would be set is if you asked for headers only where the payload was not included; in that case, that header would be added. In the direct APIs, headers are added to map values like sequence etc.

nats-base-client/kv.ts
aricart: Same as the other - so that would only trigger on a watch/history with the right option

piotrpio (approved):
LGTM

PR #479, opened 2023-02-08
[FIX] malformed status created a NaN code

aricart:
A malformed status line, the client tried to parse a space which would yield NaN - this is degenerate, but at least it doesn't have an unexpected code value

### DIFF
--- a/nats-base-client/headers.ts
+++ b/nats-base-client/headers.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 The NATS Authors
+ * Copyright 2020-2023 The NATS Authors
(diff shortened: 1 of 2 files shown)


aricart:
TBH, I think that it should be OK to have a status with no values in the header - or HPUB with 0 header information (no status), but it shouldn't be OK for a status line to have extra characters ie - if it has a status that is not just `NATS/1.0` ie `NATS/1.0 12 something` but understand that the server likely shouldn't be doing those tests. In the case of the PR above, my only discrepancy is to not provide a `NaN` for a code when the client failed to parse what it though would contain a code  and description.

derekcollison (approved):
LGTM

PR #484, opened 2023-02-21
timestamp types were incorrectly tagged as Nanos rather than strings

aricart:
[FIX] Added StoredMessage#timestamp returns the ISO message timestamp as provided by the server (`#time()` returns a `Date` object which will have lower precision)

[FIX] `StreamInfo.created` to be a `string` (was marked as `Nanos`), `StreamState#first_ts` to be a string.

Fixes https://github.com/nats-io/nats.js/issues/556


scottf (approved):
LGTM

PR #486, opened 2023-03-01
fixed duplicate kv/object store lists when unexpected stream types were included

aricart:
[FIX] Lister was filtering stream responses for objectstore and kv, but offset calculations are made as a whole. Added check of total count that lister results are not duplicated.


piotrpio (approved):
LGTM

PR #490, opened 2023-03-08
[FIX] ObjectStore meta subject entries name encoding

### LINKED ISSUE #489: Base64 padding mismatch
I have a situation where I'm trying to pull down data from an object store put there by a Go client.  Pulling this data from the cli with `--trace` shows the underlying metadata being called on `$O.web-friendly-images.M.MTIyMDE5NTQ0MDYwMzY5MDA2NjJfMDAwMTlfZnVsbA==`  however the same object being pulled from JS shows up as `$O.web-friendly-images.M.MTIyMDE5NTQ0MDYwMzY5MDA2NjJfMDAwMDFfZnVsbA`. 

This leads the js clients to see this as a 404 even though the data is available in the CLI.  It appears there is a discrepancy between clients in how padding of base64 is handled.  I couldn't find any documentation on which client is "right" though I assume the Go one is the closest to a reference implementation. 

If you need more details please don't hesitate to ask.

aricart:
[FIX] [BREAKING] [OS] previous versions of object store incorrectly base64 encoded without padding - this made javascript created entries be incompatible with other clients. The tool bin/fix-os.ts can be used to fix existing object stores - however it is important for a backup of the stream to be made prior to using it.

FIX #489

### DIFF
--- a/bin/fix-os.ts
+++ b/bin/fix-os.ts
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2023 The NATS Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


scottf (approved):
LGTM

delaneyj (approved):
LGTM 🎆

PR #492, opened 2023-03-13
[FEAT] srv metadata

aricart:
[FEAT] This adds support for metadata at the service and endpoint levels, if specified they are reported, on service info/stats.

[CHANGE] the service_test validation program will now look for metadata to be set on the service and on the endpoint. The name of the metadata/value can be anything, it requires at least one property.


ripienaar (approved):
LGTM, dont read javascript well though :)

tests/service_test.ts
```diff
@@ -745,6 +752,8 @@ Deno.test("service - cross platform service test", async () => {
     p.stderrOutput(),
   ]);
 
+  console.log(StringCodec().decode(_stdout), StringCodec().decode(stderr));
```
bruth, line 755: Log statement left over?

bruth (approved):
Other than the `console.log` statement, looks great!

tests/service_test.ts
aricart: on the test driver, let me remove that.

PR #495, opened 2023-03-16
pull consumer - consumer deleted notifications

aricart:
added ability for client to see on pull consumers to handle consumer deleted.


Jarema (approved):
LGTM

PR #498, opened 2023-04-22
[FIX] add the direct stream API to JetStreamManager interface

### LINKED ISSUE #497: TypeScript typings/interface doesn't include JetStreamManager.direct
The [`.direct`](https://github.com/nats-io/nats.deno/blob/main/nats-base-client/jsm.ts#L40) API on `JetStreamManager` is missing from the [interface](https://github.com/nats-io/nats.deno/blob/main/nats-base-client/types.ts#L766)

tothandras:
fixes #497

### DIFF
--- a/nats-base-client/types.ts
+++ b/nats-base-client/types.ts
@@ -772,6 +772,10 @@ export interface JetStreamManager {
    * JetStream API to interact with Streams
    */
   streams: StreamAPI;
+  /**
+   * JetStream Direct stream API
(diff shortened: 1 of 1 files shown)


aricart:
I actually don't want to expose this yet - you are free to use it by doing the cast, but we are in the process of changing some of the APIs for this to allow for more than one message to be retrieved, and until that ships, I rather keep this API internal.

PR #508, opened 2023-05-11
streams object

aricart:
[FEAT] added the ability to `get()` a stream name - this API returns an object that allows for easy access to `alternates()`, `best()` (best mirror as per nats-server), `getConsumer()` to get an existing consumer (or ordered consumer) associated with the stream, and `getMessage()` to retrieve a message manually. This functionality is already available via JSM or via the consumer's API; this provides more convenient access.


scottf (approved):
LGTM

PR #514, opened 2023-05-17
service changes

aricart:
Changes to match ADR - https://github.com/nats-io/nats-architecture-and-design/pull/219

Note that the ADR is not prescribing whether endpoints are added on the config, but feedback is that it is confusing to have one that is via configuration, so best just to simplify the UX of API by removing alternative mechanisms


piotrpio (approved):
LGTM!

PR #516, opened 2023-05-18
echo demo

aricart:
Example to match:
https://github.com/nats-io/natscli/blob/cb47d4fbf57b191d84c0badbeb4c763a8ab46b69/cli/micro_command.go#L71-L127

```bash
> deno run -A https://raw.githubusercontent.com/nats-io/nats.deno/echo-demo/examples/services/echo-service.ts
echo-service [--server server:port] (service | monitor | request)

Usage:
  echo-service [commands]

Available Commands:
  monitor   monitor [ping|info|status] [--name n] [--id id]
  request   make a request a service
  start     start an instance of the echo service

Flags:
  -h, --help     display echo-service's help
      --server   NATS server host:port

# start a service
> deno run -A https://raw.githubusercontent.com/nats-io/nats.deno/echo-demo/examples/services/echo-service.ts start

# make a request
> deno run -A https://raw.githubusercontent.com/nats-io/nats.deno/echo-demo/examples/services/echo-service.ts request --payload "alberto"
{ ConnectedUrl: [ "nats://demo.nats.io:4222" ] }
{
  timestamp: "Thu, 18 May 2023 20:45:46 GMT",
  request: { body: "alberto", subject: "echo" }
}

# discover
> deno run -A https://raw.githubusercontent.com/nats-io/nats.deno/echo-demo/examples/services/echo-service.ts monitor ping|stats|info

```


ripienaar (commented):
Seems good, I wonder if for demo repos we need to add all the client and management bits? May be a bit daunting and look like a ton of work, where we could have just a focus on the service for multiple languages and the cli being the client.

Then it looks nice and small and approachable?

aricart:
@ripienaar I took out all the service monitoring/request commands from the example - but I did add a `nats-services` tool that serves as an example for monitoring the services - this is not part of the demo

examples/services/echo-service.ts
```diff
@@ -0,0 +1,73 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { cli, Command, Flags } from "https://deno.land/x/cobra@v0.0.9/mod.ts";
+import { connect, headers } from "../../src/mod.ts";
+
+const root = cli({
+  use: "echo-service [--server server:port]",
```
ripienaar, line 20: https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-21.md :P

ripienaar (commented):
Yeah I like this, short and sweet, easy to pull out into a bit of doc as a guide through and starting point.

We should think a bit on standard naming conventions and config via environment vars, I might have some time this weekend to write up a go example thats 12 factor(ish) enough.  Would be great if you can mix a go and javascript one and get exact behavior.

PR #517, opened 2023-05-22
relax ordered consumer requirements

aricart:
[FIX] relaxed simplified ordered consumer to not required 2.10 unless using filtered_subjects


scottf (approved):
LGTM

PR #522, opened 2023-05-30
pub string

aricart:
[FEAT] client can now publish/request/respond with strings or Uint8Array, not that JSON still requires an encoder

[FEAT] Msg#json() and related can now specify a [reviver](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#the_reviver_parameter) function for rehydrating the JSON input.


Jarema (approved):
LGTM

PR #523, opened 2023-05-30
Changes to service info and stats

aricart:
[CHANGE] [SRV] [BREAKING] info operation now returns endpoints instead of `subjects`. Endpoints contain not only the subject for each of the endpoints but also the associated endpoint metadata

[CHANGE] [SRV] [BREAKING] removed NamedEndpointStats#metadata reference; use `EndpointInfo#metadata` instead


piotrpio (approved):
LGTM

PR #524, opened 2023-05-30
[FEAT] [JS] stream object can now delete a message - also stream info  is able to specify info options

aricart:


piotrpio (approved):
LGTM

PR #525, opened 2023-05-31
some bench enhancements

aricart:
[bench] added flag to control pendingLimit
[bench] added --rep, allowing --req/--rep to be run in different processes
[bench] changed Perf to use performance.now


scottf (approved):
LGTM

PR #526, opened 2023-06-01
[FIX] flusher timers

aricart:
When a fast producer dominates the event loop and sends more data than can fit into the outbound buffer, an inline flush happens - but on the first command sent to the server, an auto-flush is scheduled on the next event loop. The timer for this auto-schedule flush was not managed. 

In a bench-type scenario where the client is trying to send a significant amount of data, the timers accumulate, degrading the client's performance. This fix removes the scheduled timer whenever an inline flush happens, greatly improving the client's performance.

### DIFF
--- a/nats-base-client/protocol.ts
+++ b/nats-base-client/protocol.ts
@@ -155,6 +155,7 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
   servers: Servers;
   server!: ServerImpl;
   features: Features;
+  flusher?: unknown;
 
   constructor(options: ConnectionOptions, publisher: Publisher) {
     this._closed = false;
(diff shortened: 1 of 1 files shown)


piotrpio (approved):
LGTM

nats-base-client/protocol.ts
```diff
@@ -603,10 +604,17 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
...
+      this.flusher = setTimeout(() => {
         this.flushPending();
       });
     } else if (this.outbound.size() >= this.pendingLimit) {
+      // if we have a flusher, clear it - otherwise in a bench
+      // type scenario where the main loop is dominated by a publisher
+      // we create many timers.
+      if (this.flusher) {
+        clearTimeout(this.flusher);
+        this.flusher = undefined;
```
caspervonb, line 616: Nit: nullable type, so `null`? (strict null check would disallow this).

```suggestion
        this.flusher = null;
```

aricart: fixed

PR #528, opened 2023-06-01
deno socket tweaks

aricart:
[deno] changed socket read buffers 128K
[deno] removed buffer writer on connection, and instead changed to writeAll() - performance change is not noticeable, but in theory should remove some copying/sys calls


caspervonb:
> theory should remove some copying/sys calls
Would've removed a copy, since flush was on every call.

PR #530, opened 2023-06-05
Limitted microtasks

aricart:


src/deno_transport.ts
```diff
@@ -192,51 +178,87 @@ export class DenoTransport implements Transport {
...
+
+    // if we have a timer, we reached a limit of microtasks
+    // and we are waiting on a timer
+    if (this.syncTimer) {
+      // do nothing, there's a setTimeoutScheduled
+      return;
+    }
+    // if we have max number of pending microtasks
+    // schedule a timeout
+    if (this.microtasks === 1000) {
```
caspervonb, line 222: Nit: Add a const for this, tested up to a million and doesn't fail.

```suggestion
    if (this.microtasks === 1000) {
```

caspervonb (approved):
Tested extensively, lgtm (one nit, optional).

aricart:
@caspervonb  So I added a quick test printing the pending microtasks to see if there was some relation to when deno hits the limit on `Uncaught RangeError: Maximum call stack size exceeded` - and found something that doesn't make sense to me - never more than one or two microtask is queued - never. The PR is solid in performance, but I want to understand why.

As I read the code, Protocol#sendCommand() is not awaiting, so in a bench loop, it is executing - meaning that the event loop has not yet terminated, meaning that microtasks should be queued (indeed - if you print flush pending - many happen before the first microtask shows), yet this doesn't happen.

PR #533, opened 2023-06-07
chunk example

aricart:
[DOCS] example for a service that generates a requested amount of data, and then sends it in chunks, including the client making the request and processing the response


scottf (approved):
LGTM

PR #534, opened 2023-06-08
[DENO] optimization on write to just queue up

aricart:
@caspervonb this is effectively your simplification, including the changes to support `debug` and the closing promise.

### DIFF
--- a/src/deno_transport.ts
+++ b/src/deno_transport.ts
@@ -13,13 +13,12 @@
  * limitations under the License.
  */
 
-import { BufWriter } from "https://deno.land/std@0.177.0/io/mod.ts";
 import {
   Deferred,
   deferred,
   TlsOptions,
 } from "../nats-base-client/internal_mod.ts";
-import Conn = Deno.Conn;
+import { writeAll } from "https://deno.land/std@0.190.0/streams/write_all.ts";
 import {
   checkOptions,
   checkUnsupportedOption,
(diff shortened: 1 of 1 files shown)


src/deno_transport.ts
```diff
@@ -192,49 +173,41 @@ export class DenoTransport implements Transport {
...
-      .finally(() => {
-        this.sendQueue.shift();
-        this.dequeue();
+    const frame = this.frames.shift();
+    if (!frame) {
+      return;
+    }
+    this.pendingWrite = writeAll(this.conn, frame).then(() => {
+      this.pendingWrite = null;
+      queueMicrotask(() => {
```
caspervonb, line 187: This didn't have an observable effect so unsure if we should be keeping it.

```suggestion
      queueMicrotask(() => {
```

caspervonb (approved):
Benched a bit, definitively faster than main (this is on deno 1.33.x as 1.4 has major issues).

```
Repositories/nats.deno (on branch 'main') $ deno run --reload -A examples/bench.js --subject a --pub --payload 3500 --count 1000000  --pendingLimit 1024
pub 96,089 msgs/sec - [10.41 secs] ~ 320.73 MB/sec 10407/10407

Repositories/nats.deno (on branch 'write-perf') $ deno run --reload -A examples/bench.js --subject a --pub --payload 3500 --count 1000000  --pendingLimit 1024
pub 421,322 msgs/sec - [2.37 secs] ~ 1.37 GB/sec 2373.4810420000003/2373.4810420000003
```

src/deno_transport.ts
```diff
@@ -259,7 +230,10 @@ export class DenoTransport implements Transport {
         // this is a noop but gives us a place to hang
         // a close and ensure that we sent all before closing
         // we wait for the operation to fail or succeed
-        await this.enqueue(TE.encode(""));
+        this.frames.push(Empty);
+        const d = deferred<void>();
+        this.maybeWriteFrame(d);
```
caspervonb, line 235: Could just await this.pendingWrite here?

PR #545, opened 2023-06-20
[FEAT] [KV] expose stream `metadata`

aricart:


jetstream/kv.ts
```diff
@@ -917,6 +920,10 @@ export class KvStatusImpl implements KvStatus {
   get size(): number {
     return this.si.state.bytes;
   }
+
+  get metadata(): Record<string, string> | undefined {
+    return this.si.config.metadata;
```
caspervonb, line 925: Sure you don't want to default to an empty object when undefined? less cognitive overload on the caller.

caspervonb: Also looks like object store does return ?? empty.

jetstream/kv.ts
```diff
@@ -917,6 +920,10 @@ export class KvStatusImpl implements KvStatus {
   get size(): number {
     return this.si.state.bytes;
   }
+
+  get metadata(): Record<string, string> | undefined {
```
caspervonb, line 924:
```suggestion
  get metadata(): Record<string, string> {
```

jetstream/kv.ts
aricart: you are absolutely correct

caspervonb (approved):
lgtm

PR #547, opened 2023-06-22
status iterator doesn't cleanup if the iterator is stopped

### LINKED ISSUE #546: Memory leak in consumers framework API?
Me team was using `js.pullSubscribe`, but we found the new API kinda comfortable to work with and decided to update our codebase. However, we were unable to migrate the codebase without troubles.

Consider the following code snippet:

```typescript
const natsUrl = Deno.env.get("NATS_URL") ?? "";
const nc = await connect({ servers: natsUrl });
const js = nc.jetstream();
const consumer = await js.consumers.get("ENTITY", "ENTITY_CONSUMER");
while (!nc.isClosed()) {
  const msg = await consumer.next();
  msg?.nak();
  // Remove this line to see the leak occurring faster.
  await delay(1_000);
}
```

<details>
<summary>Consumer Info</summary>

image
</details>

Memory will leak dramatically making an app to fail with V8 error.

Is there any principal mistake in the code above? Thanks in advance.

aricart:
[FIX] pull adds a status listener on every call, and has no mechanism for removing it - this modifies it so that if the status iterator closes it is removed from the list

Fixes #546

### DIFF
--- a/jetstream/consumer.ts
+++ b/jetstream/consumer.ts
@@ -23,6 +23,7 @@ import {
   MsgHdrs,
   NatsError,
   QueuedIterator,
+  Status,
   Subscription,
 } from "../nats-base-client/core.ts";
 import * as core from "../nats-base-client/idleheartbeat.ts";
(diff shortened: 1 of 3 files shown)


scottf (approved):
LGTM

PR #559, opened 2023-07-22
Fixed bug in code example

y2k4life:
A variable name changes from the variable assignment when calling `consume` and the variable used in the `for` loop.


aricart (approved):
LGTM

aricart:
@y2k4life thank you for catching this and making the docs better.

PR #560, opened 2023-07-27
[FIX] ordered consumer max_redeliver set to 1, and to use mem_storage

aricart:
[FIX] ordered consumer sets redeliver to 1, and to be mem based
[TEST] ordered consumer by date turns into sequence 
[LINT] fixed some linter warnnings


Jarema (approved):
LGTM!

PR #561, opened 2023-07-27
stream/consumer info timestamps, stream configuration `

aricart:
[FEAT] stream configuration can specify the starting sequence that the stream will use (2.10.0 and better server)
[FEAT] StreamInfo and ConsumerInfo now report the UTC timestamp of when the info was created (`ts` field) (2.10.0 and better server)
[CHANGE] ConsumerInfo incorrectly typed `created` as Nanos, instead of an UTC timestamp
[FIX] ordered consumer sets redeliver to 1, and to be mem based
[TEST] ordered consumer by date turns into sequence on restarts
[LINT] fixed some linter warnnings

### DIFF
--- a/jetstream/jsapi_types.ts
+++ b/jetstream/jsapi_types.ts
@@ -84,6 +84,11 @@ export interface StreamInfo extends ApiPaged {
    * closer and faster to access.
    */
   alternates?: StreamAlternate[];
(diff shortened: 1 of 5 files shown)


Jarema (approved):
LGTM!

PR #564, opened 2023-08-10
feat(KV): Add support for watch with history

tinou98:
This adds support for a watch with history on a KV Bucket

I did not regenerate the docs (not sure how to do this, neither if it should be done in this PR)


aricart:
If you are looking for some other behaviour please reopen with additional information.

aricart:
@tinou98 ended up adding something that will do what you want:
https://github.com/nats-io/nats.deno/commit/06c37f06d2ec894813c1a696595be43ce2c4c1c4

PR #565, opened 2023-08-15
reduced latency

aricart:
nats javascript library when migrated to 2.0.0 and made to support other runtimes removed the use of `setImmediate()` which was used to flush the pending buffer once the main event loop completed. This was changed to setTimeout() executing as soon as possible. In node this had a negative performance issue, as flushing waited minimum timer resolution.


aricart:
This version is more complex and requires an opt-in to make it happen. https://github.com/nats-io/nats.deno/pull/566 achieves similar performance and all outbound buffer handling

PR #585, opened 2023-09-06
[CHANGE] require consumer/stream names to not contain spaces, newlines, carriage returns or tabs

### LINKED ISSUE #584: Clamp validation on stream/consumer names and kv/object store
### What version were you using?

- Add additional checks so that stream/consumer names that contain spaces, tabs, newlines, and carriage returns are rejected

https://github.com/nats-io/nats-architecture-and-design/issues/236

### What environment was the server running in?

any

### Is this defect reproducible?

yes

### Given the capability you are leveraging, describe your expectation?

ensure fast fails on bad names

### Given the expectation, what is the defect you are observing?

possible to create bad entries

aricart:
Fixes #584

### DIFF
--- a/jetstream/jsmconsumer_api.ts
+++ b/jetstream/jsmconsumer_api.ts
@@ -116,9 +116,16 @@ export class ConsumerAPIImpl extends BaseApiClient implements ConsumerAPI {
       throw new Error(`consumer 'name' requires server ${min}`);
     }
     if (name) {
-      const m = minValidation("name", name);
-      if (m.length) {
-        throw new Error(`consumer 'name' ${m}`);
+      try {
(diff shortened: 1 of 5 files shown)


aricart:
For KV, keys and buckets have actually even more restrictions:
https://github.com/nats-io/nats.deno/blob/b4e6de04e809766a840355df9332898cf17553b3/jetstream/kv.ts#L121

for OS, bucket names are validated: https://github.com/nats-io/nats.deno/blob/034f590f159d0613895bebb2148f57f9879bcf4e/jetstream/objectstore.ts#L61
Keys etc are base64 encoded so it doesn't matter what is provided

PR #590, opened 2023-09-15
[FEAT] allow micro to specify a queue name

aricart:
This allows the same service to be deployed under different queue groups and provide for better rtt when horizontally scaling during deployments


Jarema (approved):
LGTM

PR #596, opened 2023-09-20
os changes to update and links

aricart:
[JETSTREAM] [OS] [CHANGE] The `update()` method will now delete the previous meta for an object - previously, it reused the storage of the entry - this matched the original go client implementation.

[JETSTREAM] [OS] [CHANGE] links now always store link metadata - previously, if, on the same object store, it would duplicate the source metadata entry.

[JETSTREAM] [OS] [FIX] links of links are now rejected

[DOC] added documentation on the object store API


Jarema (approved):
LGTM

PR #597, opened 2023-09-20
[FEAT] consumer limits

aricart:
[FEAT] [JETSTREAM] added support for `consumer_limits` on stream create/update configurations - this feature enables the stream to provide a default assigned to consumers when values are not specified.


Jarema (approved):
LGTM

PR #602, opened 2023-09-29
[FEAT] sync subscription utility

aricart:
[FEAT] added a utility function `syncIterator()` to wrap an `AsyncIterables` so that elements can be extracted in a `sync` faction. Note messages and elements will continue to buffer in the iterable, so if you don't drain it or stop its source memory will grow - this is syntactic sugar, as the iterator can be had via `src[Symbol.asyncIterator]()`. This can be useful to process subscription messages one at a time; this is especially true on tests.

```javascript
const sub = nc.subscribe("foo");
nc.publish("foo");
nc.publish("foo");
const sync = syncIterator(sub);
console.log((await sync.next()).subject);
console.log((await sync.next()).subject);
```

### DIFF
--- a/jetstream/tests/consumers_test.ts
+++ b/jetstream/tests/consumers_test.ts
@@ -51,6 +51,7 @@ import {
   PullConsumerMessagesImpl,
 } from "../consumer.ts";
 import { deadline } from "../../nats-base-client/util.ts";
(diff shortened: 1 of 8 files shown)


scottf (approved):
LGTM

PR #606, opened 2023-10-06
[FEAT] exported consumer - this allows getting a Consumer without performing any JSM operations on the server

aricart:


scottf (approved):
LGTM

PR #607, opened 2023-10-06
[FEAT] added creation option to nc.jetstreamManager() to prevent checking if the JetStream API is available.

aricart:


scottf (approved):
LGTM

PR #613, opened 2023-10-16
Performance improvements

theovanderdonk:
I have found and solved 2 performance issues that together account for about 5% of processing time of my application:
1. Use of errors also in the happy flow, just for the sake of improving stack traces. Creating errors is expensive (since node/deno has to fetch the stack trace for this). Imho, recent versions of node have decent support for generating the proper stack trace from async code. So, I think the original approach is not required anymore. And if it is, it is best placed behind a feature toggle like "async-stack-traces".
2. In `request.ts`, an error is created and passed to a deferred.resolve, but when we call `cancel` from the `resolver`, the promise is already resolved or rejected, so it makes no sense to do this again. But it still is costly because the error has to be created.

### DIFF
--- a/nats-base-client/util.ts
+++ b/nats-base-client/util.ts
@@ -65,8 +65,6 @@ export interface Timeout<T> extends Promise<T> {
 }
 
 export function timeout<T>(ms: number): Timeout<T> {
-  // by generating the stack here to help identify what timed out
-  const err = NatsError.errorForCode(ErrorCode.Timeout);
   let methods;
   let timer: number;
   const p = new Promise((_resolve, reject) => {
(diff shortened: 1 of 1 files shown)


aricart:
@theovanderdonk thanks for the contribution - I will test in the different runtimes, and at worse, putting it behind the feature is a good idea - however the reason for this was runtime issues in deployments where this information was not available (typically related to doggy code that didn't handle the rejections).

theovanderdonk:
Already any results/progress on this?

aricart:
@theovanderdonk I tweaked your PR in #622  and expanded it. To disable The trace collection you have to specify the `noAsyncTraces` connection option.

aricart:
@theovanderdonk thank you very much for the contribution.

theovanderdonk:
No problem, very welcome. Thanks for the quick fix and improvements.
Any idea when this will appear in the npm package for node.js?

theovanderdonk:
And - should documentation be updated that this flag exists for top-notch performance?

PR #615, opened 2023-10-16
Consumer changes

aricart:
[FEAT] significantly more resilient simplified consumers, consumers will now jitter lookups, and in the case of ordered consumer creation of the consumer.

[FEAT] ordered consumer can now report `status()`.


Jarema (approved):
LGTM!

PR #619, opened 2023-10-17
Fix 586

### LINKED ISSUE #586: KV get fails to get republished value (message) with multiple headers
### What version were you using?

nats-server: v2.9.21
nats.ws 1.16.1
client running on Chrome browser

### What environment was the server running in?

PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian

## Steps to reproduce

I create a KV named `B`

`nats kv add B`

I create a stream named `DC_B` to subscribe to all message on the subject `A.>` and republish to `$KV.B.>`
```
nats stream add DC_B 
      --subjects='A.>'
      --republish-source='>'
      --republish-destination='$KV.B.>'
      --storage=memory 
      --replicas=1 
      --retention=limits 
      --discard=old 
      --max-msgs=1 
      --max-msgs-per-subject=-1 
      --max-bytes=-1 
      --max-age=-1 
      --max-msg-size=-1 
      --dupe-window="2m0s" 
      --allow-rollup 
      --no-deny-delete 
      --no-deny-purge
```

I publish a message on `A`

`nats pub "A.tomato" "hello"`

As a sanity check I also explicitly `put` a value directly into `B` to compare

`nats kv put B 'A.potato' "yo"`

It does get capture by the stream `DC_B` and does get republished to the KV bucket `B` correctly
```
-> A.tomato "hello"
    -> KV B
        -> A.tomato: "hello"
```

Watching with `nats sub '>'` while this is happening I can see the message that goes to `B` contains multiple headers
```
[#48] Received on "_INBOX.Olz80ftB4Yj0cBmlJhobQO.XOE2JA36"
Nats-Stream: DC_B
Nats-Stream: KV_B
Nats-Subject: A.tomato                 <------ this
Nats-Subject: $KV.B.A.tomato           <------ and this
Nats-Sequence: 2
Nats-Sequence: 2
Nats-Last-Sequence: 1
Nats-Time-Stamp: 2023-09-02T14:48:06.891798Z

hello

[#43] Received on "_INBOX.qppVDoINpVj2TabJ60qInw.LhKzBUHR"
Nats-Stream: KV_B
Nats-Subject: $KV.B.A.potato
Nats-Sequence: 3
Nats-Time-Stamp: 2023-09-02T14:52:48.235748Z

yo
```

Using the CLI I can get the value using the key `A.tomato` from `B` as expected 👍 
```
nats kv get B "A.tomato"

hello
```
This also works using the Python client correctly. 👍 

Here is the bug 🐛  in the JS implementation 👇 
```javascript
const B = await js.views.kv("B")
let tomato = await B.get("A.tomato")
let potato = await B.get("A.potato")
```

I get `tomato` is `null` 😞  and and `potato` is correctly `"yo"`

The only difference is that `A.tomato` was republished into the KV and has multiple headers.

I tracked it down to the `Bucket` class's `get` function here https://github.com/nats-io/nats.deno/blob/db885246875d3c1d6401e1f686a9a0570ac20e9b/jetstream/kv.ts#L528
This equality fails
```javascript
      if (ke.key !== ek) {
```

And it fails due to the incorrect header retrieval in the`length` getter in the `KvStoredEntryImpl` class here
https://github.com/nats-io/nats.deno/blob/db885246875d3c1d6401e1f686a9a0570ac20e9b/jetstream/kv.ts#L984C5
```javascript
get length() {
    const slen = this.sm.header.get(JsHeaders.MessageSizeHdr) || ""
    if (slen !== "") {
        return parseInt(slen, 10)
    }
    return this.sm.data.length
}
```

The call to `this.sm.header.get` randomly (?) returns one of the headers for `Nats-Subject` so it doesn't match the subject equality check.

The value is correctly being retrieved but it is rejected and thrown out.

### Given the capability you are leveraging, describe your expectation?

I expect 
```javascript
const B = await js.views.kv("B")
let tomato = await B.get("A.tomato")
```
to return the value stored.

### Given the expectation, what is the defect you are observing?

Multiple headers (specifically `Nats-Subject`) in a KV key messages cause the value to be irretrievable by the JS implementation while Python and CLI (assuming Go as well) work.

aricart:
[FEAT] added `last()` to MsgHdrs to retrieve the last header value if multiple headers share the same name.

[FIX] changed direct header implementation to read last header value instead of first. If message was republished, the last header value will be the correct one for the message

Fix #586

### DIFF
--- a/jetstream/jsm.ts
+++ b/jetstream/jsm.ts
@@ -100,11 +100,11 @@ export class DirectMsgImpl implements DirectMsg {
   }
 
   get subject(): string {
-    return this.header.get(DirectMsgHeaders.Subject);
+    return this.header.last(DirectMsgHeaders.Subject);
   }
 
   get seq(): number {
-    const v = this.header.get(DirectMsgHeaders.Sequence);
+    const v = this.header.last(DirectMsgHeaders.Sequence);
     return typeof v === "string" ? parseInt(v) : 0;
   }
 
@@ -113,11 +113,11 @@ export class DirectMsgImpl implements DirectMsg {
   }
 
   get timestamp(): string {
-    return this.header.get(DirectMsgHeaders.TimeStamp);
+    return this.header.last(DirectMsgHeaders.TimeStamp);
   }
 
   get stream(): string {
-    return this.header.get(DirectMsgHeaders.Stream);
+    return this.header.last(DirectMsgHeaders.Stream);
   }
 
   json<T = unknown>(reviver?: ReviverFn): T {
--- a/jetstream/tests/consumers_test.ts
+++ b/jetstream/tests/consumers_test.ts
@@ -19,13 +19,7 @@ import {
   assertRejects,
   assertStringIncludes,
 } from "https://deno.land/std@0.200.0/assert/mod.ts";
-import {
-  deferred,
-  Empty,
-  NatsConnection,
-  nuid,
-  StringCodec,
-} from "../../nats-base-client/mod.ts";
+import { deferred, Empty, StringCodec } from "../../nats-base-client/mod.ts";
 import {
   AckPolicy,
   Consumer,
--- a/jetstream/tests/kv_test.ts
+++ b/jetstream/tests/kv_test.ts
@@ -1894,3 +1894,41 @@ Deno.test("kv - watch history no deletes", async () => {
 
   await cleanup(ns, nc);
 });
+
+Deno.test("kv - republish header handling", async () => {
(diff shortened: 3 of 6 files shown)


Jarema (approved):
LGTM!

PR #630, opened 2023-10-30
[FEAT] added `reason` to JsMsg#term - this feature requires 2.11.0

aricart:
[TEST] moved util to different test file otherwise some tests run multiple times


scottf (approved):
LGTM

PR #637, opened 2023-11-15
Test cluster tool fixes

mtmk:
Path and tmp folder fixes fpr Windows.
Also signal handling fixes for Windows and Linux.

### DIFF
--- a/tests/helpers/cluster.ts
+++ b/tests/helpers/cluster.ts
@@ -46,6 +46,11 @@ if (argv.h || argv.help) {
   Deno.exit(0);
 }
 
+// Windows can't handle signals to stop/start chaos
+if (Deno.build.os === "windows" && !argv["chaos"]) {
+  console.log("use --chaos flag to restart cluster servers randomly")
+}
+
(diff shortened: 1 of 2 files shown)


tests/helpers/cluster.ts
```diff
@@ -140,28 +146,36 @@ function restart(cluster: NatsServer[]) {
 }
 
 function waitForStop(): void {
+  const isWindows = Deno.build.os === "windows";
+  const isLinux = Deno.build.os === "linux";
```
aricart, line 150: lets just test for windows, the rest is assumed to be some sort of unix, otherwise we'll fail and add more complexity for other things that will run

aricart (changes requested):

tests/helpers/cluster.ts
mtmk: This is only for signal handling on Linux swapping `ctrl-T` with `ctrl-Z` since `SIGINFO` doesn't work on Linux. It'd be nice to have the option to run on Linux as well. BTW macOS should not be affected.

PR #641, opened 2023-12-01
prevent consumer deleted from stopping iterator

aricart:
[FIX] changed consumer so that a server consumer_deleted notification doesn't stop the iterator. The consumer will notify the condition, and continue to re-attempt to retrieve messages.


ripienaar (approved):
LGTM

PR #656, opened 2024-03-01
[FEAT] [KV] delete and purge at seq

aricart:
delete and purge now take an optional sequence. If the current key is not at the specified revision, the delete/purge fails


scottf (approved):
LGTM

PR #662, opened 2024-03-12
allow internal kv watch consumer to use filtered consumer api

aricart:
[FIX] changed KV ordered consumers created by watch, to set `name` if supported by the server. This enables the client to be able to use the newer consumer APIs which can then be clamped down with permissions on filter

[FIX] removed unused JSM reference in the Views implementation

### DIFF
--- a/jetstream/tests/kv_test.ts
+++ b/jetstream/tests/kv_test.ts
@@ -2048,13 +2048,6 @@ Deno.test("kv - watcher will name and filter", async () => {
 
   const js = nc.jetstream();
   const kv = await js.views.kv("A");
-  await Promise.all([
-    kv.put("a.b", "2"),
-    kv.put("a.cc", "3"),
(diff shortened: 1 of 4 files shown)


jetstream/tests/kv_test.ts
```diff
@@ -2036,3 +2037,34 @@ Deno.test("kv - bind no info", async () => {
...
+    kv.put("b.a", "3"),
+    kv.put("b.b", "3"),
+    kv.put("b.c", "3"),
+  ]);
+
+  const sub = syncIterator(nc.subscribe("$JS.API.>"));
+  const iter = await kv.watch({ key: "a.>" });
+
+  const m = await sub.next();
+  console.log(m?.subject);
```
ramonberrutti, line 2063: console.log

aricart: @ramonberrutti thanks Ramon!!

PR #665, opened 2024-03-15
enhancements to downstream iterator use to determine if the iterator failed

aricart:
[CHANGE] iterator close()/closed() now return a promise that could resolve to an Error, this simplifies upstream cases where it is necessary to know if there was an error while processing the iterator.


scottf (approved):
LGTM

PR #666, opened 2024-03-19
jetstream also support jsm options

ramonberrutti:
Extend jetstream options with jetstreamManager options


nats-base-client/nats.ts
```diff
@@ -505,7 +505,7 @@ export class NatsConnectionImpl implements NatsConnection {
   }
 
   jetstream(
-    opts: JetStreamOptions = {},
+    opts: JetStreamOptions | JetStreamManagerImpl = {},
```
aricart, line 508: this doesn't look right - JetstreamManagerImpl?

nats-base-client/core.ts
```diff
@@ -651,7 +651,7 @@ export interface NatsConnection {
    *
    * @param opts
    */
-  jetstream(opts?: JetStreamOptions): JetStreamClient;
+  jetstream(opts?: JetStreamOptions | JetStreamManagerOptions): JetStreamClient;
```
aricart, line 654: Should this be `JetStreamOptions & JetstreamManagerOptions` - this would allow all options be be specified.

aricart (changes requested):

nats-base-client/core.ts
aricart: Doh - the current type for JetStreamManagerOptions extends JetStreamOptions, so this is a good change.

aricart (approved):
LGTM

aricart:
@ramonberrutti thank you!

nats-base-client/nats.ts
ramonberrutti: you are right

PR #668, opened 2024-03-22
[FEAT] added a bind option to the consume, fetch, next - note this option is not supported in ordered consumers.

aricart:


jetstream/consumer.ts
```diff
@@ -139,7 +139,7 @@ export type Expires = {
 
 export type Bind = {
   /**
-   * If set to false the client will not try to check on its consumer by issuing consumer info
+   * If set to true the client will not try to check on its consumer by issuing consumer info
    * requests. This means that the client may not report consumer not found, etc., and will simply
```
mtmk, line 143: would it help to add a sentence about consumer info requests potentially being expensive on the server?

scottf (approved):
LGTM

jetstream/consumer.ts
aricart: Not sure that is appropriate - specially since that is a server issue that should be addressed there.

PR #669, opened 2024-03-26
[TEST] added a test to verify that newly expanded options for jetstream/jetstream manager honors checkAPI option

aricart:


piotrpio (approved):
LGTM!

PR #671, opened 2024-03-26
[FIX] failing test and bad options

ramonberrutti:
This fixed an error introduced here: https://github.com/nats-io/nats.deno/pull/666


aricart (commented):
Actually the test here is wrong, because the options are overridden in the wrong order. I have a PR for this.

aricart:
This is the correct fix

https://github.com/nats-io/nats.deno/pull/672

aricart (commented):
Will use https://github.com/nats-io/nats.deno/pull/672 instead

PR #675, opened 2024-03-27
Fix incorrect type in implementation of `jetstream()` api - it is correct in the interface.

aricart:
[FIX] jetstream() api implementation incorrectly typed options as JetStreamManagerImpl - should have been JetStreamManagerOptions - this is only in the implementation, the types are correct on the interface.


ramonberrutti (approved):
LGTM

mtmk (approved):
LGTM

PR #676, opened 2024-03-28
ackAck() always returning false

aricart:
[FIX] JsMsg.ackAck() was not waiting for a proper response from the server. Now it will return false, if the reply subject is not set, or the message was already ack'ed. And will reject if the there's a problem with the response.


mtmk (approved):
LGTM

PR #682, opened 2024-04-08
[FEAT] `handshake_first` support

aricart:
[FEAT] [CORE] connect to `handshake_first` configured NATS servers specify `{tls: {handshakeFirst: true}}` connection option. Note that the server must match the configuration option, and will only accept with a matching option.


scottf (approved):
LGTM

PR #686, opened 2024-04-15
[FEAT] [CORE] added the ability to publish by providing a Msg argument

aricart:
This effectively uses a message as a set of parameters. The client will use the provided message as the source for subject, data, reply and headers. Note that this will work with any argument that satisfies the message interface, but it is not of interest to most applications.


scottf (approved):
LGTM

PR #687, opened 2024-04-19
[CHANGE] moved `nanos()` and `millis()` to nats-base-client as they are referenced by the base library

aricart:
[CI] [BUMP] nats-server


scottf (approved):
LGTM

PR #689, opened 2024-04-22
[FIX] Kv Watcher reliability during server restarts

### LINKED ISSUE #688: KV watch functionality fails when NATS server restarts
### Observed behavior

When watching for KV updates and the NATS server restarts, the client reconnects successfully but API throws the error:
```
{
   code: 400,
   err_code: 10013,
   description: 'consumer name already in use'
 }
```

Debug logs show that it tries to reconnect twice, with the same request data, when the watch method has been called.

### Expected behavior

It should reconnect and continue receiving KV updates 

### Server and client version

nats-server 2.10.14
nats.js 2.22.0 

### Host environment

Unix OS

### Steps to reproduce

From the [KV example](https://github.com/nats-io/nats.deno/blob/main/examples/legacy_js/12_kv.ts), the code to reproduce looks like:

```
const nats = require('./lib/src/mod.js');
const nc = await nats.connect();
const sc = nats.StringCodec();
const js = nc.jetstream();
const kv = await js.views.kv("testing");
await kv.create("hello.world", sc.encode("hi"));

const watch = await kv.watch();
(async () => {
  for await (const e of watch) {
    // do something with the change
    console.log(
      `watch: ${e.key}: ${e.operation} ${e.value ? sc.decode(e.value) : ""}`,
    );
  }
})().then();
```

1. Run the example app
2. Restart NATS server
3. The client reconnects but the watch functionality fails with the response:

`{"type":"io.nats.jetstream.api.v1.consumer_create_response","error":{"code":400,"err_code":10013,"description":"consumer name already in use"}}`

Any ideas? Thanks!

aricart:
[FIX] [JS] [KV] watcher reset due to a server restart could fail because the server would report the consumer as existing

[FIX] [JS] [KV] watcher reset if successful, could still get heartbeat missing reports which would cause the ordered consumer to reset when it didn't need to.

Fix #688

### DIFF
--- a/jetstream/jsclient.ts
+++ b/jetstream/jsclient.ts
@@ -98,6 +98,7 @@ import {
   PullOptions,
   ReplayPolicy,
 } from "./jsapi_types.ts";
+import { nuid } from "../nats-base-client/nuid.ts";
 
 export enum PubHeaders {
   MsgIdHdr = "Nats-Msg-Id",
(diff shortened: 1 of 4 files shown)


scottf (approved):
LGTM

PR #694, opened 2024-05-02
[FIX] legacy order consumer subscription leak

### LINKED ISSUE #691: JetStream KV gets consumer stuck in a cluster node
### Observed behavior

The KV consumer subscription remains active without interest after the client reconnects to another available cluster node.

Freshly, I had discussed the possibility of nats-server issue (see [5352](https://github.com/nats-io/nats-server/issues/5352)) and it has been raised that it might be a client bug when keeping interest on the old subscription.

### Expected behavior

Consumers should be unsubscribed/removed if the client has reconnected to another available cluster node and has no interest on it.

### Server and client version

- nats-server 2.10.14
- nats.js 2.23.0

### Host environment

- Unix OS
- Docker
- Cluster mode

### Steps to reproduce

Considering that nats-server has 3 cluster servers running

1. Run an app to watch for JetStream KV updates with a cluster mode setup, Node.js example:
```
const nc = await nats.connect({
    servers: ['localhost:4111','localhost:4222','localhost:4333'],
  });
  const sc = nats.StringCodec();
  const js = nc.jetstream();
  const kv = await js.views.kv("testing");
  const jsm = await nc.jetstreamManager();
  const stream = await jsm.streams.get('KV_testing');
  const { config } = await stream.info();

  await jsm.streams.update(config.name, {
    ...config,
    num_replicas: 3
  });

  const watch = await kv.watch();
  (async () => {
    for await (const e of watch) {
      // do something with the change
      console.log(
        `watch: ${e.key}: ${e.operation} ${e.value ? sc.decode(e.value) : ""}`,
      );
    }
  })();

  kv.put('hello.word', sc.encode("hi"));
```

2. Check which cluster has been chosen: nats consumer report KV_testing
3. Force container restart/node fail situation, so the client will connect to another available cluster node
4. Check the consumer report again: nats consumer report KV_testing
5. Output shows two consumers instead of one expected

aricart:
[FIX] fixed an issue with ordered consumer where resetting the consumer didn't send the UNSUB, creating a situation where if there was no disconnect from the client, two subscriptions would be active on the server.

FIX #691

### DIFF
--- a/jetstream/jsclient.ts
+++ b/jetstream/jsclient.ts
@@ -795,6 +795,8 @@ export class JetStreamSubscriptionImpl extends TypedSubscription<JsMsg>
     this.js._request(subj, req, { retries: -1 })
       .then((v) => {
         const ci = v as ConsumerInfo;
+        const jinfo = this.sub.info as JetStreamSubscriptionInfo;
+        jinfo.last = ci;
         this.info!.config = ci.config;
         this.info!.name = ci.name;
       })
--- a/nats-base-client/protocol.ts
+++ b/nats-base-client/protocol.ts
@@ -995,6 +995,7 @@ export class ProtocolHandler implements Dispatcher<ParserEvent> {
     if (!s || this.isClosed()) {
       return;
     }
+    this.unsub(s);
     s.subject = subject;
     this.subscriptions.resub(s);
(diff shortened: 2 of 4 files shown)


scottf (approved):
LGTM

PR #697, opened 2024-05-09
[FEAT] [JS] [CONSUMERS] added `name_prefix` option to ordered consumer that allows user-provided prefix to generated consumer names

### LINKED ISSUE #696: Add ability to prefix an Ordered Consumer Name
### Proposed change

Add an optional  `prefix` argument to `OrderedConsumerOptions` to allow the creation of ordered consumers with names beginning with the given prefix

### Use case

The motivation behind this is to determine who / what is creating order consumers via an identifier that can be given when creating an ordered consumer. Without it, a random name for a consumer is created. It's difficult to tell where it came from and what is creating them in a production environment. 

### Contribution

_No response_

aricart:
Fix #696


ripienaar (approved):
LGTM

PR #701, opened 2024-05-30
[FIX] fixed an issue where permission errors related queue subscriptions was not properly notified to the client

aricart:


scottf (approved):
LGTM

PR #703, opened 2024-06-13
[FEAT] added connect option `noResolve` which disables name resolution on the client

aricart:
FIX [https://github.com/nats-io/nats.js/issues/625


scottf (approved):
LGTM

PR #704, opened 2024-06-13
[FIX] simpler way to noResolve

aricart:


aricart:
Fixes https://github.com/nats-io/nats.js/issues/625

scottf (approved):
LGTM

PR #708, opened 2024-06-19
Honor JS timeout in ackAck() and also allow to override

aricart:
[FIX] [JS] JsMsg#ackAck() didn't provide a way to customize the timeout. Added optional option with timeout for ackAck(), and also defaulted the timeout to match the JSClient's timeout (as provided by the JetStreamOptions)


scottf (approved):
LGTM

PR #710, opened 2024-06-20
[FEAT] [KV] added support for listing keys with multiple filters

aricart:
[FEAT] [KV] added support for listing keys with multiple filters (client already supported a single filter, this enables the possibility of specifying additional keys)


scottf (approved):
LGTM

PR #712, opened 2024-06-21
Package moderniation

sachaw:
Hi, A lot has change in the Deno ecosystem as of late and I was wandering if you were open to some changes.

Namely the advent of JSR, Moving to JSR would allow for a single cross-platform package (preferably with Deno as the primary runtime)
Thus far, I have moved all dependencies to an import map (in `deno.json`)
There appears to be a lot of simplification possible with the current "build" process (can likely do away with all of the bundling)

What I would like to know is if there are any major objections with perusing the goal of providing a single cross-platform client library? I'm more than happy to invest some significant maintenance time into this and the associated libraries

Thanks.

### DIFF
--- a/bin/bundle.ts
+++ b/bin/bundle.ts
@@ -12,8 +12,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { cli } from "https://deno.land/x/cobra@v0.0.9/mod.ts";
-import { bundle } from "https://deno.land/x/emit@0.26.0/mod.ts";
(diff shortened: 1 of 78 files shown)


aricart:
I have all of this under work already! - in the [split-js-core ](https://github.com/nats-io/nats.deno/tree/split-core-js) branch. Will incorporate the fix for the avj you put there. Thank you.

sachaw:
Excellent, I'll definitely be contributing more in the near future.

PR #717, opened 2024-07-09
Map `opts.replicas` to `sc.num_replicas` in Object Store init

SalvaChiLlo:
The library interface, defined by the `ObjectStoreOptions` type, currently expects a parameter named `replicas` to specify the number of replicas for the object store. However, internally, the correct parameter to be used is `num_replicas`. To align with the existing initialization method of KV, the initialization method of OB should also map `opts.replicas` to `sc.num_replicas`. This ensures consistency and proper utilization of the `replicas` parameter within the object store configuration.


aricart (approved):
LGTM - will add a test and simplify as KV to just assign opts.replicas and let the server deal with the absence of the value.

aricart:
@SalvaChiLlo thank you for spotting this.

PR #721, opened 2024-07-29
fix(kv): fixed an issue where kv sources were not properly processed

aricart:
- added a required transform to the source
- added the kv subject when sources are present


jnmoyne (approved):
LGTM

PR #727, opened 2024-08-22
nano fixes

aricart:
doc(core,jetstream): jsdoc for fields that are in nanoseconds, but where not specified in the description - jsdoc erases the type alias and marks as number

fix(services): `processing_time` was reported in millis instead of nanos


scottf (approved):
LGTM

PR #745, opened 2025-02-07
[BREAKING] [FIX] replace sha256 library used by the client for objectstore entries and migration tool

### LINKED ISSUE #746: [OS] can register objects with a digest that is incorrect
### Observed behavior

The sha256 lib used by the client can generate wrong hashes depending on the size of the file

### Expected behavior

should always match `openssl sha256 somefile`

### Server and client version

all previous objectstore versions

### Host environment

_No response_

### Steps to reproduce

_No response_

aricart:
ObjectStore implementations prior to this one used a sha256 library that sporadically generated incorrect digests for larger objects. These digests prevent other tools from accessing objects `put` by the JavaScript ObjectStore implementation. Included in the fix is a migration tool that walks a specified objectstore and calculates hashes, if found to be a match for the bad-hashes, it updates the hash in the metadata entry for the object.

Fixes #746


scottf (approved):
LGTM

Issue #751, opened 2025-03-13
OrderedPullConsumerImpl mutating options object


HNicolas:
### Observed behavior

Some methods from the `OrderedPullConsumerImpl` are mutating the provided options object resulting in possible unexpected behavior.
For instance successive calls to `const iterator = await consumer.fetch(options)` will result in no message being passed to the iterator.

### Expected behavior

The options object should not be mutated.

### Server and client version

Using nats node npm package v2.29.2.
Nats server docker image `nats:2.9.16-alpine`.

### Host environment

Running nats in docker and node v22.13.0.

### Steps to reproduce

Code to reproduce the issue:
```ts
import { connect, DeliverPolicy } from "nats";

const nats = await connect({
  servers: "127.0.0.1:4222"
});
const js = nats.jetstream();
const consumer = await js.consumers.get("todos", { deliver_policy: DeliverPolicy.New });

const options = { max_messages: 10 };
while (true) {
  const iter = await consumer.fetch(options);
  for await (const message of iter) {
    console.log("consumed message", message.seq);
  }
}
```
After the first call to fetch, iterator will not return new messages.

The following code does not have the issue:
```ts
import { connect, DeliverPolicy } from "nats";

const nats = await connect({
  servers: "127.0.0.1:4222"
});
const js = nats.jetstream();
const consumer = await js.consumers.get("todos", { deliver_policy: DeliverPolicy.New });

while (true) {
  const iter = await consumer.fetch({ max_messages: 10 });
  for await (const message of iter) {
    console.log("consumed message", message.seq);
  }
}
```

aricart:
This is currently fixed in the 3.x series

See https://github.com/nats-io/nats.js/pull/224

PR #752, opened 2025-03-13
[FIX] do not mutate options in consumer methods

HNicolas:
#751

### DIFF
--- a/jetstream/consumer.ts
+++ b/jetstream/consumer.ts
@@ -314,7 +314,7 @@ export class PullConsumerMessagesImpl extends QueuedIteratorImpl<JsMsg>
     super();
     this.consumer = c;
 
-    const copts = opts as ConsumeOptions;
+    const copts = { ...opts } as ConsumeOptions;
 
     this.opts = this.parseOptions(opts, refilling);
     this.callback = copts.callback || null;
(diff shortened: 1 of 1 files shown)


aricart (changes requested):
@HNicolas thank you - could you please add a test case.

jetstream/consumer.ts
```diff
@@ -314,7 +314,7 @@ export class PullConsumerMessagesImpl extends QueuedIteratorImpl<JsMsg>
     super();
     this.consumer = c;
 
-    const copts = opts as ConsumeOptions;
+    const copts = { ...opts } as ConsumeOptions;
```
aricart, line 317: would it be better to do `structuredClone()` in case some of those structures ever become objects, etc.

aricart: some of those will take a callback - so cannot use structuredClone()...

HNicolas:
I will see to add a test in the week.

aricart:
I have released the 3.0 libs, and this feature is available there - see https://github.com/nats-io/nats.js/releases/tag/v3.0.0

https://github.com/nats-io/nats.js/commit/a504a35248633c050267f9d23cf8b9e81f4b4737

